PR #4, opened 2022-10-24
Add Android Project Structure

Jacob4649:
This PR will initialize an Android Project Structure, it will replace the default gradle files, and add an example activity, instrumented test, and unit test.


Jacob4649:
There seems to be an issue with the replacement gradle config not working with the Autograding workflow tests.

Jacob4649:
The tests don't pass so we're deleting the tests XD

PR #11, opened 2022-10-27
move model to java8

Jacob4649:


Jacob4649:
Changes the model module's Gradle build config to use Java 8 instead of Java 7. This is to keep consistent with the rest of the project. Java 8 is used instead of Java 11, because Android only added Java 11 support in 2021, and thus, applications intended to run on legacy Android devices, which do not have the newest version of Android, must use Java 8.

PR #27, opened 2022-10-31
UserInfo entities

BullDF:
Finished creating UserInfo, ProviderInfo, and ClientInfo classes

### DIFF
--- a/model/src/main/java/userinfo/ClientInfo.java
+++ b/model/src/main/java/userinfo/ClientInfo.java
@@ -1,0 +1,44 @@
+package userinfo;
+
+/**
+ * Stores the information of a client.
+ */
+public class ClientInfo extends UserInfo {
+    private String firstName;
+    private String lastName;
+
+    /**
+     * Constructs a ClientInfo object.
+     *
+     * @param address: the address of this client
+     * @param phoneNumber: the phone number of this client
+     * @param firstName: the first name of this client
+     * @param lastName: the last name of this client
+     */
+    public ClientInfo(String address, String phoneNumber, String firstName, String lastName) {
+        super(address, phoneNumber);
+        this.firstName = firstName;
+        this.lastName = lastName;
+    }
+
+    public String getFirstName() {
+        return firstName;
+    }
+
+    public void setFirstName(String firstName) {
+        this.firstName = firstName;
+    }
+
+    public String getLastName() {
+        return lastName;
+    }
+
+    public void setLastName(String lastName) {
+        this.lastName = lastName;
+    }
+
+    @Override
+    public String getType() {
+        return "CLIENT";
+    }
+}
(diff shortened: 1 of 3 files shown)


Lenzs (dismissed):

Jacob4649:
In the future, more descriptive commit messages than 'Files created' might be useful. Keep in mind that commits will not show up in history in the context of the branch they were on before they were merged, so commit messages should make sense without the context of the current branch. 

A good way to think about what constitutes a good commit message is to fill in the sentence 'when merged, this commit will...' for the current commit. The second half of the sentence makes for a great commit message.

E.X.

Consider a commit adding an `ExampleClass` class.

Filling in the sentence as explained above leaves 'when merged this commit will add the `ExampleClass` class'.

A good commit message would be just the second half of the sentence above, 'add the `ExampleClass` class'.

Jacob4649 (changes requested):
Instead of a package called userinfo specifically for `UserInfo` and subclasses, it might make more sense to have a user package containing all `User` related entities (i.e. `User`, `UserInfo`, etc...)

model/src/main/java/userinfo/UserInfo.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public void setPhoneNumber(String phoneNumber) {
+        this.phoneNumber = phoneNumber;
+    }
+
+    /**
+     * Gets the type of the user.
+     * @return "CLIENT" or "PROVIDER" depending on the type of this user
+     */
+    public abstract String getType();
```
Jacob4649, line 41: Instead of returning a `String` here, it might make more sense to return an enum of some kind as laid out in the CRC model. A `UserType` enum with `CLIENT` and `PROVIDER` fields could be created.

model/src/main/java/userinfo/ProviderInfo.java
```diff
@@ -0,0 +1,53 @@
...
+        this.certified = false;
+    }
+
+    public String getPracticeName() {
+        return practiceName;
+    }
+
+    public void setPracticeName(String practiceName) {
+        this.practiceName = practiceName;
+    }
```
Jacob4649, line 29: These should be documented

model/src/main/java/userinfo/ClientInfo.java
```diff
@@ -0,0 +1,44 @@
...
+        this.firstName = firstName;
+    }
+
+    public String getLastName() {
+        return lastName;
+    }
+
+    public void setLastName(String lastName) {
+        this.lastName = lastName;
+    }
```
Jacob4649, line 38: These should be documented, it doesn't have to be super detailed, for getters, generally you only need to fill in the @returns.

Jacob4649 (changes requested):
In the future, make sure to write unit tests for every single PR. Each developer is responsible for writing tests for their own code, so make sure to add some unit tests for these new entity classes. They're definitely pretty simple, so there isn't really much to test, but it's still good practice.

Additionally, instead of a package called userinfo specifically for UserInfo and subclasses, it might make more sense to have a user package containing all User related entities (i.e. User, UserInfo, etc...)

Jacob4649:
Force pushes to fix history

PR #31, opened 2022-11-04
add User entity

Lam207:


model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
+package com.javahelp.model.userEntityTest;
+import com.javahelp.model.user.UserInfo;
+import com.javahelp.model.user.ClientUserInfo;
+import com.javahelp.model.user.UserType;
+import com.javahelp.model.user.ProviderUserInfo;
+
+public class User {
+    /**
+     * A class for a user entity
+     */
```
Jacob4649, line 10: The convention in Java is to have Javadocs above the method/class/field they pertain to rather than below. For example, this section should look like

```java
/**
 * A class for a user entity
 */
public class User {
```

model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private String stringID;
+    private UserInfo userInfo;
```
Jacob4649, line 13: We may also want a username or email field here. This would be unique, like ID, but unlike ID, would not be immutable.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * A constructor for the current User Entity
+     *
+     * @param stringID - the ID of the account
+     * @param userInfo - The account information.
+     */
+
```
Jacob4649, line 21: This whitespace shouldn't be here, Javadocs should be directly above the item they refer to.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
...
+        this.userInfo = userInfo;
+    }
+
+    /**
+     * Set the StringID for the user
+     *
+     * @param stringID - the updated ID of the account
+     *                 userInfo - the updated information of the user
+     */
+
```
Jacob4649, line 33: See above comment on whitespace.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
...
+    public void setStringID(String stringID) {
+        this.stringID = stringID;
+    }
+
+    public void setUserInfo(UserInfo userInfo) { this.userInfo = userInfo; }
```
Jacob4649, line 38: While it does compile, this method should probably be expanded into several lines:

```java
public void setUserInfo(UserInfo userInfo) {
    this.userInfo = userInfo;
}
```

Additionally, this method needs documentation.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Getters of all the instance variables of the current user.
+     *
+     * @return StringID - the current ID
+     * userInfo - the current information of the user
+     */
+
+    public String getStringID() {
+        return stringID;
+    }
```
Jacob4649, line 49: Whitespace comment, and also Javadocs should describe a specific method, rather than a family of methods, etc... This comment describes all getters in the class, it should instead specifically describe the `getStringID` getter.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public UserInfo getUserInfo() {
+        return userInfo;
+    }
```
Jacob4649, line 53: Needs docs

model/src/main/java/com/javahelp/model/userEntityTest/UserPassword.java
```diff
@@ -0,0 +1,99 @@
...
+     * salt - a current salt of a user
+     * hash - the current and hashed password of the user
+     * user - the user object
+     */
+
+    private byte[] salt;
+
+    private byte[] hash;
+
+    private User user;
```
Jacob4649, line 18: In my opinion, the `UserPassword` class should not contain a reference to the `User` the password belongs to. The purpose of the class should more be to bundle together a user's salt and hash, and also to disconnect this information from the `User` object, so this information is not needlessly loaded into memory.

model/src/main/java/com/javahelp/model/userEntityTest/UserPassword.java
```diff
@@ -0,0 +1,99 @@
...
+        int position = 0;
+        for (byte element: salt){
+            representation[position] = element;
+            position++;
+        }
+        for (byte element: hash){
+            representation[position] = element;
+            position++;
+        }
+        return representation;
```
Jacob4649, line 83: See `System.arraycopy`

model/src/test/java/com/javahelp/model/userEntityTest/UserEntityTest.java
```diff
@@ -0,0 +1,76 @@
...
+public class UserEntityTest {
+
+    /** These are tests for User Entity
+     *
+     * --- Attributes ----
+     * user1 - The first user object
+     * user2 - the second user object
+     * userinfo1 - the information of the first user
+     * userinfo2 - the information of the second user
+     */
```
Jacob4649, line 21: Reiterating the point from `User`, the convention with Javadocs is to put them above the class, with no whitespace between the bottom of the comment, and the start of the class. Additionally, the conventional way to add Javadocs for fields is to place separate comments above the fields. For private fields, this is not necessary.

model/src/test/java/com/javahelp/model/userEntityTest/UserPasswordTest.java
```diff
@@ -0,0 +1,160 @@
...
+     * saltOne - the current salt for the first user
+     * saltTwo - the current salt for the second user
+     * passWord1 - the current given password of the first user
+     * passWord2 - the current given password of the second user
+     *
+     * user1Pass - an object represents for the first user's password
+     * user2Pass - an object represents for the second user's password
+     * user1 - the first user
+     * user2 - the second user
+     */
```
Jacob4649, line 31: See the earlier comments on Javadocs

model/src/test/java/com/javahelp/model/userEntityTest/UserPasswordTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private Random rand;
```
Jacob4649, line 33: The `Random` instance does not really need to be stored in the class here

Jacob4649 (changes requested):

Jacob4649:
Also, I totally forgot to mention this in my review, but the package should be the same as the package for the `UserInfo` entities. It should be `com.javahelp.model.user` rather than `com.javahelp.model.userEntityTest`. Also note that it is conventional to not use camelCase or PascalCase in Java package names.

model/src/test/java/com/javahelp/model/userEntityTest/UserPasswordTest.java
Jacob4649: Also just a note for the future, you don't need to put `git stage . && git commit -m "commit message" && git push` in your commit messages. It seems like you might be typing this into the message space in Android Studio, but these are console commands. The message space is just for the commit message, and Android Studio does the rest automatically.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
Lam207: Yeah, I think username (or the user's first name and last name) as well as their email are stored inside UserInfo userInfo. Is it necessary to extract them out explicitly, though?

model/src/main/java/com/javahelp/model/userEntityTest/User.java
Lam207: Thanks!

model/src/main/java/com/javahelp/model/userEntityTest/User.java
Lam207: Got it! Yeah, this was short but not really concise. I prioritized looking shorter, so I grouped them way too much.

model/src/test/java/com/javahelp/model/userEntityTest/UserPasswordTest.java
Lam207: Oh I see, I'll do ``git commit -m "this is a commit message"`` next time

Jacob4649: You can just write the commit message in the commit message box in Android Studio and it'll do all the command line stuff automatically. I wouldn't worry about any git commands.

model/src/main/java/com/javahelp/model/userEntityTest/User.java
Jacob4649: My line of thought is that the username and/or email really has nothing to do with the personal information of each client or provider, and is really more administrative. If it's something like a username, I think it probably belongs in user, but for I guess maybe for something like an email it makes more sense in UserInfo.

model/src/test/java/com/javahelp/model/usertest/ClientUserInfoTest.java
```diff
@@ -1,4 +1,4 @@
-package com.javahelp.model.userTest;
+package com.javahelp.model.usertest;
```
Jacob4649, line 1: package should just be `com.javahelp.model.user` on these. It is conventional to keep tests in same package as the classes they pertain to.

model/src/test/java/com/javahelp/model/usertest/ProviderUserInfoTest.java
```diff
@@ -1,4 +1,4 @@
-package com.javahelp.model.userTest;
+package com.javahelp.model.usertest;
```
Jacob4649, line 1: See test package comment

model/src/test/java/com/javahelp/model/usertest/UserEntityTest.java
```diff
@@ -1,25 +1,25 @@
-package com.javahelp.model.userEntityTest;
+package com.javahelp.model.usertest;
```
Jacob4649, line 1: See test package comment

model/src/test/java/com/javahelp/model/usertest/UserEntityTest.java
```diff
@@ -1,25 +1,25 @@
...
 
-    /** These are tests for User Entity
-     *
+    /**
      * --- Attributes ----
      * user1 - The first user object
      * user2 - the second user object
      * userinfo1 - the information of the first user
      * userinfo2 - the information of the second user
      */
```
Jacob4649, line 22: This comment is kinda unconventional for Java. Instead of one big comment for all fields, the convention is to write a single Javadoc for each public facing field. In this case, it's not really necessary.

model/src/test/java/com/javahelp/model/usertest/UserPasswordTest.java
```diff
@@ -29,7 +29,6 @@ public class UserPasswordTest {
      * user1 - the first user
      * user2 - the second user
      */
-
     private Random rand;
```
Jacob4649, line 32: I believe this can be local, and does not need to be a field of the class

model/src/main/java/com/javahelp/model/user/UserPassword.java
```diff
@@ -1,63 +1,62 @@
...
      * @param hash - The already-hashed password
-     * @param user - the current user account
      */
-
-    public UserPassword(User user, byte[] salt, byte[] hash){
-        this.user = user;
+    public UserPassword(byte[] salt, byte[] hash){
         this.salt = salt;
         this.hash = hash;
     }
```
Jacob4649, line 27: Now that I think about it, we may also want a constructor that takes in a `String` and a salt, and hashes the string to create a new `UserPassword`. This may be outside of the scope of this PR for now, so perhaps it should be left alone for the time being.

Jacob4649: Actually now that I think about it, a constructor such as the one that currently exists will most likely never be used. We will need a constructor that hashes a `String` eventually. Maybe for now, it's best to leave this without any constructor.

model/src/main/java/com/javahelp/model/user/UserPassword.java
```diff
@@ -1,63 +1,62 @@
...
         return salt;
     }
 
-
-    /** The setters of the current attributes. Please note that only current salt and
-     * password can be set, while the current user always stays the same
-     * in this UserPassword entity!
+    /** Set current hashed password to the user.
+     * Please note that only hashed password can be set, while the current user always
+     * stays the same in this UserPassword entity!
```
Jacob4649, line 47: This is true, but really the `UserPassword` just doesn't have any reference to a `User` so it's more that `UserPassword` is user-agnostic than that the user is always the same

model/src/main/java/com/javahelp/model/user/UserPassword.java
```diff
@@ -1,63 +1,62 @@
...
      * @param hash - the currently hashed password
-     *        salt - the current salt
      */
-
     public void setHash(byte[] hash) {
         this.hash = hash;
     }
 
+    /** Set the current salt.
+     * Please note that the consistency in user object applies to salt, too.
```
Jacob4649, line 56: I'm unsure what this means

model/src/main/java/com/javahelp/model/user/UserPassword.java
```diff
@@ -70,16 +69,13 @@ public void setSalt(byte[] salt) {
...
-        byte[] representation = new byte[salt.length + hash.length];
-        int position = 0;
-        for (byte element: salt){
-            representation[position] = element;
-            position++;
-        }
-        for (byte element: hash){
-            representation[position] = element;
-            position++;
+        if (salt.length == 0 || hash.length == 0) {
```
Jacob4649, line 72: Eventually, with a different constructor, this will no longer be possible; however, for now it makes sense.

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,84 @@
...
+     * Set the StringID for the user
+     *
+     * @param userInfo - an updated ID of the account
+     */
+    public void setUserInfo(UserInfo userInfo) { this.userInfo = userInfo; }
+
+    /**
+     * Set the username for the user
+     *
+     * @param username - an updated username of the account
```
Jacob4649, line 53: You might want to check [this](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html) out, just to get an idea of how Javadocs are meant to look

Jacob4649 (changes requested):

model/src/test/java/com/javahelp/model/user/UserPasswordTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * A testing file for UserPassword entity
+ */
+public class UserPasswordTest {
+
+    /**
+     * The current salt for the first user
+     */
+    private byte[] saltOne;
```
Jacob4649, line 20: Just to emphasize, this is the correct format for Javadocs, but in a situation like this (on a private field, especially one in a test class) they are not necessarily required. This is the right format though.

This isn't a request for changes, just a note for the future.

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,84 @@
...
+public class User {
+
+    /**
+     * --- Attributes---
+     * stringID - the ID of the user account
+     * userInfo - all information of the user. It varies based on Client/Provider
+     */
+    private String stringID;
+    private UserInfo userInfo;
+    private String username;
```
Jacob4649, line 19: Same thing as before with Javadocs formatting. For private fields, Javadocs are not necessary. In this case, you could either turn these into comments on the fields themselves, or remove the comment entirely.

Jacob4649 (changes requested):
Overall, looks really good! Definitely almost ready to merge IMHO.

Jacob4649 (dismissed):
Looks great!

yuanshu777 (dismissed):
I like it, the codes are very clear

Lam207:
> Looks great!

Thanks very much for your help :)

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,86 @@
...
+    public void setStringID(String stringID) {
+        this.stringID = stringID;
+    }
+
+    /**
+     * Set the StringID for the user
+     *
+     * @param userInfo an updated ID of the account
+     */
+    public void setUserInfo(UserInfo userInfo) { this.userInfo = userInfo; }
```
BullDF, line 50: I don't know if this is GitHub's auto contraction but this line looks incorrectly formatted. We should expand the curly brackets into multiple lines like all other functions.

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * A class for a user entity
+ */
+public class User {
+
+    /**
+     * The ID of the user account (randomly generated and immutable)
+     */
+    private String stringID;
```
BullDF, line 11: I think if this is immutable then should we make it a final attribute (e.g. private final String stringID)?

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,86 @@
...
+        this.userInfo = userInfo;
+        this.username = username;
+    }
+
+    /**
+     * Set the StringID for the user
+     *
+     * @param stringID an updated ID of the account
+     */
+    public void setStringID(String stringID) {
```
BullDF, line 41: If we cannot change the stringID, we probably don't need a setter for this attribute.

model/src/main/java/com/javahelp/model/user/UserPassword.java
```diff
@@ -0,0 +1,91 @@
+package com.javahelp.model.user;
+
+import java.util.Base64;
+
+/**
+ * A class for the password of a user entity
+ */
+public class UserPassword {
```
BullDF, line 8: I admit that I am not too familiar with salt sand hash and things but, do we need any common attributes between the User class and the UserPassword class in order to match a UserPassword instance with the specific user? For example, do we need an "ID" attribute in the UserPassword class so that we know this password instance is for the user of that ID?

BullDF (changes requested):
Thanks Lam for your work. This looks pretty good. I have added some comments in your codes! Please check them out!

BullDF (commented):
Thanks Lam for your work. This looks pretty good. I have added some comments in your codes! Please check them out!

BullDF (approved):
It makes more sense now, thanks!

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,79 @@
...
+     * The username of the current user
+     */
+    private String username;
+
+    /**
+     * A constructor for the current User Entity
+     *
+     * @param stringID the ID of the account
+     * @param userInfo the account information
+     * @param username the username of the account (randomly generated)
```
BullDF, line 28: Why is username randomly generated?

BullDF, line 28: Why is username randomly generated though? Should it be created by the user?

BullDF (changes requested):

BullDF (dismissed):
Nice work Lam!

BullDF (dismissed):

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * A class for a user entity
+ */
+public class User {
+
+    /**
+     * The ID of the user account (randomly generated and immutable)
+     */
+    private final String stringID;
```
Jacob4649, line 11: this shouldn't be final, according to our spec right now, it needs to be reassigned by the frontend when fetching a newly created user.

Jacob4649 (changes requested):

model/src/main/java/com/javahelp/model/user/User.java
Jacob4649: It also needs a setter, I think it already has a getter, but it needs a setter too (the setter just won't be used very often)

model/src/main/java/com/javahelp/model/user/UserPassword.java
Jacob4649: I was thinking that could more be handled by the API gateway. I.E. you pass in a user id and get the associated `UserPassword`. Instead of storing the associated ID in the `UserPassword`, this detail can be handled by the database, and by implementations of the API gateway

model/src/main/java/com/javahelp/model/user/User.java
Jacob4649: This is only immutable in the sense that once assigned, and entered into the database, it is never changed by the user. There are still scenarios where it needs to be changed in the code, especially according to the plan in our CRC cards for creating new users. The CRC specification currently details a create method that accepts a `User` object, creates a `User` object matching the provided one in the database, assigns a random ID to the newly created user, and mutates the passed `User` object to have an ID matching its new ID in the database. To do this, we need the `stringID` field to not be final.

Jacob4649 (approved):
Looks good to me

model/src/main/java/com/javahelp/model/user/User.java
```diff
@@ -0,0 +1,89 @@
+package com.javahelp.model.user;
+
+/**
+ * A class for a user entity
+ */
+public class User {
+
+    /**
+     * The ID of the user account (randomly generated and immutable)
```
BullDF, line 9: I think we should change the javadoc here as well as to keep it clear that it's immutable ONLY when it's assigned and saved to the database

BullDF (commented):
I think we should change the attribute javadoc to more clearer as well similar to the constructor javadoc

Jacob4649 (approved):

BullDF (approved):

yuanshu777 (approved):

yuanshu777:
add user entity

PR #33, opened 2022-11-07
Token entity and Token Testing added

Enkhzul001:
I have added token entity. Please take your time and review it. Thank you.

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -10,3 +10,6 @@ java {
 dependencies {
     testImplementation 'junit:junit:4.13.2'
 }
+dependencies {
+    implementation 'joda-time:joda-time:2.9.4'
+}
--- a/model/src/main/java/com/javahelp/model/token/Token.java
+++ b/model/src/main/java/com/javahelp/model/token/Token.java
@@ -1,0 +1,123 @@
+package com.javahelp.model.token;
+
+import com.javahelp.model.user.User;
+import org.joda.time.DateTime;
+import java.security.SecureRandom;
+
+/**
+ * Representation of token entity
+ */
+
+public class Token {
+    /**
+     * Token requested by the user
+     */
+    private String token;
+
+    /**
+     * Unique id of the token.
+     */
+    private String id;
+
+    /**
+     * Expiry date of the token.
+     */
+    private DateTime expiry;
+
+    /**
+     * Issued date of the token.
+     */
+    private DateTime issued;
+
+    /**
+     * User who requested the token.
+     */
+    private User user;
+
+    /**
+     * This holds any additional information about the toekn.
+     */
+    private String tag;
+
+
+    /**
+     * Creates a new {@link Token}
+     * @param tag {@link String} of this {@link Token}
+     * @param user {@link User} who requested this {@link Token}
+     */
+    public Token(String tag, User user){
+        // Creating a token using functions from SecureRandom
+
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[20];
+        random.nextBytes(bytes);
+        this.token = bytes.toString();
+
+        this.issued = new DateTime(DateTime.now());
+        this.expiry = this.issued.plusDays(365/2);
+        this.user = user;
+        this.tag = tag;
+    }
+
+    /**
+     *
+     * @return unique {@link String} token for this {@link Token}
+     */
+    public String getToken(){
+        return this.token;
+    }
+
+    /**
+     *
+     * @return {@link String} tag for this {@link Token}
+     */
+    public String getTag(){
+        return this.tag;
+    }
+
+    /**
+     *
+     * @return unique {@link String} id for this {@link Token}
+     */
+    public String getId(){
+        return this.id;
+    }
+
+    /**
+     *
+     * @return {@link String} issued date for this {@link Token}
+     */
+    public DateTime getIssuedDate(){ return this.issued;}
+
+    /**
+     *
+     * @return {@link String} expiry date for this {@link Token}
+     */
+    public DateTime getExpiryDate(){
+        return this.expiry;
+    }
+
+    /**
+     * Setting up the token for this {@link Token}
+     */
+    public void setToken(String t){
+        this.token = t;
+    }
+
+    /**
+     * Setting up the issue date for the token.
+     * @param d1 {@link DateTime} of this {@link Token}
+     */
+    public void setIssuedDate(DateTime d1){
+        this.issued = d1;
+    }
+
+    /**
+     * Settingup the expiry date for the token.
+     * @param d2 {@link DateTime} of this {@link Token}
+     */
+    public void setExpiryDate(DateTime d2){
+        this.expiry = d2;
+    }
+
+}
--- a/model/src/test/java/com/javahelp/model/token/TokenTest.java
+++ b/model/src/test/java/com/javahelp/model/token/TokenTest.java
@@ -1,0 +1,49 @@
+package com.javahelp.model.token;
+
+import com.javahelp.model.user.ClientUserInfo;
+import com.javahelp.model.user.User;
+import org.joda.time.*;
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Unit tests for {@link Token}
+ */
+public class TokenTest {
+
+    // Setting up the user and token
+    ClientUserInfo info = new ClientUserInfo("user1@example.com", "New York",
+            "111","David", "Smith");
+    User user1 = new User("123", info, "user1");
+    Token token = new Token("This is Test token", user1);
+
+    @Test(timeout = 50)
+    public void getToken(){
+        token.setToken("[!jkl123");
+        String expected_token = "[!jkl123";
+        assertEquals(expected_token,token.getToken());
+    }
+
+    @Test(timeout = 50)
+    public void getTag(){
+        assertEquals("This is Test token", token.getTag());
+    }
+
+    @Test(timeout = 50)
+    public void getIssuedDate(){
+        DateTime issuedDate = token.getIssuedDate();
+        token.setIssuedDate(issuedDate);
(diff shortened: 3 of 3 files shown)


model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,123 @@
...
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[20];
+        random.nextBytes(bytes);
+        this.token = bytes.toString();
+
+        this.issued = new DateTime(DateTime.now());
+        this.expiry = this.issued.plusDays(365/2);
+        this.user = user;
+        this.tag = tag;
+    }
```
Jacob4649, line 60: We need a way to assign a specific token string to a `Token` object, it would be best if this was possible from the constructor. This constructor should take an additional `String` argument that sets the `Token` private field, instead of randomly generating a byte array to put in the private field.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,123 @@
...
+     * @param tag {@link String} of this {@link Token}
+     * @param user {@link User} who requested this {@link Token}
+     */
+    public Token(String tag, User user){
+        // Creating a token using functions from SecureRandom
+
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[20];
+        random.nextBytes(bytes);
+        this.token = bytes.toString();
```
Jacob4649, line 54: This may not do what you are thinking. You might want to read into the string representations of arrays in Java, and how they are represented by default. This string is not reflective of the contents of the array, and is also not really random.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,123 @@
...
+public class Token {
+    /**
+     * Token requested by the user
+     */
+    private String token;
+
+    /**
+     * Unique id of the token.
+     */
+    private String id;
```
Jacob4649, line 20: We may want to get rid of the id field, and simply leave the token field.

Jacob4649: The argument for this is that the id is somewhat redundant, as the token string is already created at the same time as the token is, and is already immutable. We can just use the token string as the id.

model/build.gradle
```diff
@@ -9,4 +9,7 @@ java {
 
 dependencies {
     testImplementation 'junit:junit:4.13.2'
+}
+dependencies {
+    implementation 'joda-time:joda-time:2.9.4'
 }
```
Jacob4649, line 15: Adding Joda is ok if there are no alternatives; however, I believe for all the applications in this PR either, `java.util.Date` or `java.time` work as well or better. Using these instead would allow us to keep the same functionality, while not needing to add Joda as a dependency. I believe this could be a good design decision.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,123 @@
...
+    public Token(String tag, User user){
+        // Creating a token using functions from SecureRandom
+
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[20];
+        random.nextBytes(bytes);
+        this.token = bytes.toString();
+
+        this.issued = new DateTime(DateTime.now());
+        this.expiry = this.issued.plusDays(365/2);
```
Jacob4649, line 57: We need the option to set both these fields from the constructor. We may want some tokens to expire sooner or later than others (i.e. User is signed on permanently, vs signed on for single session). We will also want to be able to set the issued date to any arbitrary date, as a `Token` object may not always be created on the date it was issued. Consider a token issued 2 weeks ago, but used to login in the present. When the backend authenticates this token, it will create a `Token` object, and will need to create one with a issued date of two weeks prior.

model/src/test/java/com/javahelp/model/token/TokenTest.java
```diff
@@ -0,0 +1,49 @@
...
+    // Setting up the user and token
+    ClientUserInfo info = new ClientUserInfo("user1@example.com", "New York",
+            "111","David", "Smith");
+    User user1 = new User("123", info, "user1");
+    Token token = new Token("This is Test token", user1);
+
+    @Test(timeout = 50)
+    public void getToken(){
+        token.setToken("[!jkl123");
+        String expected_token = "[!jkl123";
```
Jacob4649, line 25: make sure to keep Java naming conventions in mind, use camelCase instead of snake_case

Jacob4649 (approved):
Looking pretty good, just a few small things

Jacob4649 (changes requested):
Looking good, just a few small things

model/src/main/java/com/javahelp/model/token/Token.java
Lam207: Yeah, it shows the memory address of the array I think. Is the usage of StringBuilder to append elements helpful here?

model/src/main/java/com/javahelp/model/token/Token.java
Lam207: Great catch, one question though, should the Database of Token be responsible for telling when was that token created before and how long more it will be valid (if true)?

Then we might need these parameters from the Database, I guess.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Creates a new {@link Token}
+     * @param tag {@link String} of this {@link Token}
+     * @param user {@link User} who requested this {@link Token}
+     */
+    public Token(String tag, User user){
+        // Creating a token using functions from SecureRandom
+
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[20];
```
Lam207, line 52: A small comment on Java notation, is this like in ``C`` syntax? 

Because usually I see in Java, we do ``byte[] bytes = new byte[20];`` instead.

model/src/test/java/com/javahelp/model/token/TokenTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Test(timeout = 50)
+    public void getIssuedDate(){
+        DateTime issuedDate = token.getIssuedDate();
+        token.setIssuedDate(issuedDate);
+        DateTime expected_date = token.getIssuedDate();
+        assertEquals(expected_date, issuedDate);
+    }
+
+    @Test(timeout = 50)
+    public void getExpiryDate(){
```
Lam207, line 43: My alternative (and pretty dummy) approach to this is to create a new and different expiredDate first but not yet set it to the token attribute yet, and compared them for the first time. It should be notEquals for the first time, of course. 

Then, I set the new expiredDate to the attribute and compare them again, they should be Equal at this time.

model/src/main/java/com/javahelp/model/token/Token.java
Jacob4649: Yuh exactly, and idk I mean we shouldn't be building the ID inside the token class, that's a job for the implementation of the gateway to the database

model/src/main/java/com/javahelp/model/token/Token.java
Jacob4649: Yeah, that's exactly right

model/src/main/java/com/javahelp/model/token/Token.java
Jacob4649: Ye you can put it after either the type or the name, but it's probably more conventional to put it after the type

model/build.gradle
Jacob4649: also minimizing dependencies is important for keeping the size of our function package low for the backend. If it gets too big, we'll have to switch to using a container image.

model/src/main/java/com/javahelp/model/token/Token.java
Enkhzul001: You are right. I forgot to remove the id.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,79 @@
+package com.javahelp.model.token;
+
+import com.javahelp.model.user.User;
+
+import java.time.*;
+import java.security.SecureRandom;
```
Jacob4649, line 6: Might wanna clean up these imports but tbh i don't think it's a huge deal we can also do a general imports, cleanup, and formatting PR in the future.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @return {@link LocalDate} issued date for this {@link Token}
+     */
+    public LocalDate getIssuedDate(){ return this.issued;}
+
+    /**
+     *
+     * @return {@link LocalDate expiry date for this {@link Token}
+     */
+    public LocalDate getExpiryDate(){return this.expiry;}
```
Jacob4649, line 78: These should all be split across multiple lines

model/src/main/java/com/javahelp/model/token/Token.java
Jacob4649: If you're changing other stuff you might as well do these too though, you can just hit Ctrl + Alt + O to automatically clean them.

model/src/test/java/com/javahelp/model/token/TokenTest.java
```diff
@@ -0,0 +1,45 @@
+package com.javahelp.model.token;
+
+import com.javahelp.model.user.ClientUserInfo;
+import com.javahelp.model.user.User;
+import java.time.*;
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+import java.time.format.DateTimeFormatter;
```
Jacob4649, line 9: Clean imports again

Jacob4649 (approved):
Looks really good, I commented a few changes, but they're just formatting things

Lam207 (approved):
Looks good to me

yuanshu777 (approved):

PR #51, opened 2022-11-10
Setup DynamoDB Client

Jacob4649:
Setup the now configured DynamoDB to work with the AWS Lambda functions created by the Serverless library for this application. Configure the backend module to use the AWS Java SDK to connect to DynamoDB.

PR #56, opened 2022-11-11
Token gateway

### LINKED ISSUE #42: [Backend | Login & Auth API] Token Database Gateway Interface

Jacob4649:
Closes #42. Branched off #51 and #46, should only be merged after both.

### DIFF
--- a/model/src/main/java/com/javahelp/model/token/Token.java
+++ b/model/src/main/java/com/javahelp/model/token/Token.java
@@ -17,12 +17,12 @@ public class Token {
     /**
      * Expiry date of the token.
      */
-    private LocalDate expiry;
+    private Instant expiry;
 
     /**
      * Issued date of the token.
      */
-    private LocalDate issued;
+    private Instant issued;
 
     /**
      * ID of {@link User} for {@link Token}.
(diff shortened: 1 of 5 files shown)


model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -18,17 +17,17 @@ public class Token {
...
      * Expiry date of the token.
      */
-    private LocalDate expiry;
+    private Instant expiry;
 
     /**
      * Issued date of the token.
      */
-    private LocalDate issued;
+    private Instant issued;
```
Enkhzul001, line 25: What is an instant?

Enkhzul001 (approved):
Looks good. Just few questions.

yuanshu777 (approved):
LOOKS GOOD TO ME

model/src/main/java/com/javahelp/model/token/Token.java
Jacob4649: Um its like a LocalDate but not connected to any timezone

Jacob4649: or i mean a LocalDate isn't connected to a timezone but it does assume that there's some associated timezone implicitly

Jacob4649: Also LocalDate is just for dates, and doesn't do times. But basically, LocalDate represents like a day on a Calendar. But the problem is, we may have people in multiple timezones. We could still get around that (using LocalDateTime to allow for tracking time) by converting all times to a specific time zone before we interact with them, but it's easier to use something like an Instant, which instead of representing a calendar date, represents a specific moment in time, and is the same across timezones.

Lam207 (approved):
This refinement makes it even more usable across many different regions with the use of ``Instant``!

PR #64, opened 2022-11-13
Client reg UI

Enkhzul001:
I created the layout for client registration activity and added scroll view so the client can scroll down and fill their info.

### DIFF
--- a/.idea/assetWizardSettings.xml
+++ b/.idea/assetWizardSettings.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="WizardSettings">
+    <option name="children">
+      <map>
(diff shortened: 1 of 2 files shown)


Jacob4649 (changes requested):
It looks like this PR may contain old commits, redundant commits, and commits meant for other branches. It might be good to remove these.

.idea/assetWizardSettings.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Jacob4649, line 1: This file probably should not be commited

Jacob4649 (changes requested):

yuanshu777 (approved):

yuanshu777:
looks good to me

PR #69, opened 2022-11-14
[Backend] Account Deletion Manager

BullDF:
The account deletion manager, part of the account deletion use case.


backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteManager.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    protected DeleteResult delete(String userID, UserPassword userPassword) {
+        User user = iUserStore.read(userID);
+
+        if (verify(userID, userPassword)) {
+            iUserStore.delete(userID);
+            return new DeleteResult(true, user);
+        }
+
+        return new DeleteResult(false, user);
```
Enkhzul001, line 59: If the user is deleted, why do we keep the user in DeleteResult? Shouldn't DeleteResult return only success or does keeping the user in DeleteResult useful for us?

BullDF: I think it will be useful for later if we want to display any message in the frontend such as "User with {userID} has been deleted" or anything like that.

Enkhzul001 (dismissed):
Looks good to me.

BullDF:
Closed for now. Further revises will be made this afternoon.

BullDF:
The verify password feature in DeleteManager has been deleted because this can be combined with the login use case.

BullDF:
I only changed DeleteManager and DeleteResult and the corresponding test file but I don't know why other classes appeared here. Maybe because I opened and copied some of the codes. Ignore those files. Those have not been changed.

Jacob4649:
I'm removing the extra commits.

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteManager.java
```diff
@@ -0,0 +1,38 @@
+package com.javahelp.backend.accountdeletion;
```
Jacob4649, line 1: You might want to move all of these into a package like `com.javahelp.backend.domain.user.delete`

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteManager.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.javahelp.backend.data.IUserStore;
+import com.javahelp.model.user.User;
+import com.javahelp.model.user.UserPassword;
+
+/**
+ * A class for managing an account deletion action.
+ */
+class DeleteManager {
```
Jacob4649, line 10: We should be consistent with our use case class naming conventions, this is called manager, I just pushed a PR that used interactor instead. The CRC cards say manager, so it might make sense to use that and rename any interactors to be consistent, but we might also want to have this discussion with the whole team present, just to keep everyone up to date and get everyone's input!

Jacob4649, line 10: This class should be public, as it will be the entry point for the use case, and must be accessed from outside the use case

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteManager.java
```diff
@@ -0,0 +1,38 @@
...
+     * The {@link IUserStore} to use.
+     */
+    private final IUserStore userStore;
+
+    /**
+     * Constructs a {@link DeleteManager} instance.
+     *
+     * @param userStore: the {@link IUserStore} used to store users.
+     */
+    protected DeleteManager(IUserStore userStore) {
```
Jacob4649, line 22: Similar to the class, the constructor must also be public for the same reasons.

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteResult.java
```diff
@@ -0,0 +1,37 @@
...
+    private final String successMessage;
+
+    /**
+     * Constructs a {@link DeleteResult} instance for a successful account deletion.
+     *
+     * @param user: the deleted {@link User}.
+     */
+    protected DeleteResult(User user) {
+        this.user = user;
+        successMessage = "Account deletion successful";
```
Jacob4649, line 20: rather than a success message, it might make more sense to have an error message. You could have two constructors, one that takes in an error message for failures, and one that takes in a `User` for successes. You could also add a getter like `isSuccess` or something that returns whether the error message is null.

Jacob4649 (changes requested):
Overall, this looks really good

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteResult.java
BullDF: I thought about this, but I wrote it this way because the deletion is made under the circumstance that the user is logged in and there shouldn't be any error in this process. The only error I could think of is if the password doesn't match but since that is tackled by the login use case, I think we will be fine here.

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteManager.java
BullDF: Should I make the delete method public as well?

BullDF: > Should I make the delete method public as well?

I think so.

backend/src/main/java/com/javahelp/backend/accountdeletion/DeleteResult.java
Jacob4649: That's true, but we also want to be able to inform the end user in cases where something goes wrong with the database, or the string id sent doesn't exist in the first place, or anything else goes wrong. You can also imagine some scenarios where a user is logged in on two instances of the frontend, and deletes their account on one. They may then try to delete their account from the second instance. In this case, we may want to either return an error, or respond with a cached response based on some idempotency key provided with the request. If we do want to return an error, we need a way to do so.

BullDF:
Since I couldn't run the tests locally due to AWS credentials issues, I couldn't add a timeout time for the tests because I don't know how long they take to run (Sorry Lam).

Jacob4649 (approved):
Looks good!

Enkhzul001 (approved):
good work!

Lam207 (approved):
Another good way of design and handling exceptions of deleting a non-existent users 👍

PR #73, opened 2022-11-14
User login UI

yuanshu777:
Added the Login UI as our launch page for now
Added some colours code to use


build.gradle
```diff
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
-    id 'com.android.application' version '7.2.2' apply false
-    id 'com.android.library' version '7.2.2' apply false
+    id 'com.android.application' version '7.3.1' apply false
+    id 'com.android.library' version '7.3.1' apply false
 }
```
Jacob4649, line 6: we probably want to avoid updating the Gradle files

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Mon Oct 24 10:52:31 EDT 2022
 distributionBase=GRADLE_USER_HOME
```
Jacob4649, line 2: We probably also want to avoid updating this

Jacob4649 (changes requested):
Looks really good, just a few small changes with the gradle files

build.gradle
yuanshu777: solved

app/src/main/AndroidManifest.xml
```diff
@@ -13,6 +13,20 @@
...
+        <activity
+            android:name=".frontend.FrontPageActivity"
+            android:exported="false">
+            <meta-data
+                android:name="android.app.lib_name"
+                android:value="" />
+        </activity>
         <activity
             android:name=".frontend.MainActivity"
             android:exported="true">
```
Enkhzul001, line 32: What is happening here?

Jacob4649 (changes requested):
Looks good to me, but a more verbose name for `PRegActivity` might be appropriate. Something like `ProviderRegistrationActivity` (provided that's what it is meant to be).

app/src/main/AndroidManifest.xml
yuanshu777: I think for each activity, you have to declare them in the AndroidManifest, and the PRehActivity and FronPageActivity are activities that we want to link to the Mainactivity

yuanshu777:
> Looks good to me, but a more verbose name for `PRegActivity` might be appropriate. Something like `ProviderRegistrationActivity` (provided that's what it is meant to be).

I know, I thought it would be too long. But since PR isn't appropriate, I will change it. Thank you

Jacob4649 (approved):
Looks good!

BullDF (approved):
Not too familiar with this UI thing, but the graphics look good (seen during tutorial).

Enkhzul001 (approved):
Overall, looks great! I will use some of the images and colors to match the style on my UI.

PR #75, opened 2022-11-14
Auth token use case

Lenzs:


backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
```diff
@@ -0,0 +1,37 @@
...
+    private void setAuthenticate(User desiredUser, Token token) {
+        this.authenticate = new TokenAuthResult(desiredUser, token);
+    }
+
+    /**
+     * compares the input user and token and returns the result of the comparison
+     *
+     * @return TokenAuthResult object with the results of the comparison
+     */
+    private TokenAuthResult authenticate(User desiredUser, Token token) {
```
BullDF, line 33: Should we set this method to public or even protected so other classes can get this result out of this manager?

model/src/main/java/userinfo/ClientInfo.java
```diff
@@ -0,0 +1,44 @@
+package userinfo;
```
BullDF, line 1: I think we have put all the UserInfo classes and subclasses in the user package as well. You might want to move these files before we merge

BullDF (changes requested):
Overall looking good Bill. I will also suggest to write test cases for these classes to ensure they work.

model/src/main/java/userinfo/ClientInfo.java
Lenzs: i actually didnt do anything with these classes, so im not sure why they are included in my branch/PR. my best guess is that i had set up my branch incorrectly but i have no idea.

should i still make changes to this file?

backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
```diff
@@ -0,0 +1,37 @@
+package com.javahelp.backend;
+
+import com.javahelp.backend.data.ITokenStore;
+import com.javahelp.backend.data.IUserStore;
+import com.javahelp.model.token.Token;
+import com.javahelp.model.user.User;
+
+public class TokenAuthManager {
+    private IUserStore userStore;
+    private ITokenStore tokenStore;
```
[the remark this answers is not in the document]
Lenzs: thanks for the feedback! ill make these changes later tonight

Jacob4649 (changes requested):
It looks like this branch may have some commits you don't want included in this PR, let me know if you want help removing these.

model/src/main/java/userinfo/ClientInfo.java
Jacob4649: Yeah it does seem like the branch might not be set up correctly

backend/src/main/java/com/javahelp/backend/TokenAuthResult.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.javahelp.model.token.Token;
+import com.javahelp.model.user.User;
+
+import org.graalvm.compiler.api.replacements.Snippet;
+
+public class TokenAuthResult {
+    private boolean authenticated;
+    private User user;
+    private Token token;
```
BullDF, line 11: I am thinking if it will be better to set user and token (and even authenticated?) to final as well, since they are associated with only one user and the corresponding token? (Or even errorMessage to be final as well? Since if this authentication fails, the user has to initiate another login request and it will be another `TokenAuthResult` generated.) Just some considerations.

backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
```diff
@@ -0,0 +1,37 @@
...
+    public void setAuthenticate(User desiredUser, Token token) {
+        this.authenticate = new TokenAuthResult(desiredUser, token);
+    }
+
+    /**
+     * compares the input user and token and returns the result of the comparison
+     *
+     * @return TokenAuthResult object with the results of the comparison
+     */
+    public TokenAuthResult authenticate(User desiredUser, Token token) {
```
BullDF, line 33: Can this method be combined with `setAuthenticate()`? Since it's just one line of code in `setAuthenticate`. If you do combine, would it be better if we can just return the instance of `TokenAuthResult` directly instead of setting it as an attribute of the class and then returning it?

BullDF (commented):
Overall pretty good! Thanks for adding test cases!

backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
Lenzs: Yeah I can definitely combine the two methods, I think we could just return an instance of TokenAuthResult but the current version is how it's done in the CRC cards so I'm not sure if it'll come into play later

backend/src/main/java/com/javahelp/backend/TokenAuthResult.java
```diff
@@ -0,0 +1,36 @@
...
+    private String errorMessage;
+
+    private String ID;
```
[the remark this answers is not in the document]
Lenzs: ahh this one is my bad, I added this yesterday to help with some bugging but forgot to delete it afterwards

backend/src/main/java/com/javahelp/backend/TokenAuthResult.java
```diff
@@ -0,0 +1,36 @@
...
+        this.user = user;
+        this.token = token;
+        if (authenticated) {
+            this.errorMessage = null;
+        } else {
+            this.errorMessage = "Token Authentication failed";
+        }
+    }
+
+    public Boolean getAuthenticated() { return this.authenticated; }
```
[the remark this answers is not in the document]
Lenzs: yea you're right this was a typo

model/src/main/java/userinfo/ClientInfo.java
Lam207: For me it shows `package com.javahelp.model.user;`

BullDF (approved):

backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
Jacob4649: I think it would probably be better to keep the `TokenAuthResult` as a local field, and not as a field in the class.

backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
```diff
@@ -0,0 +1,33 @@
+package com.javahelp.backend;
```
Jacob4649, line 1: As a use case, this should probably go somewhere like `com.javahelp.backend.domain.user`, etc...

backend/src/main/java/com/javahelp/backend/TokenAuthManager.java
```diff
@@ -0,0 +1,33 @@
...
+        this.userStore = userStore;
+        this.tokenStore = tokenStore;
+    }
+
+    /**
+     * compares the input user and token and returns the result of the comparison
+     *
+     * @return TokenAuthResult object with the results of the comparison
+     */
+    public TokenAuthResult authenticate(User desiredUser, Token token) {
```
Jacob4649, line 29: This method should probably take in a `String desiredUserID` and a `String token`, and use the stores provided in the constructor to resolve both into objects, before comparing and producing a result.

backend/src/main/java/com/javahelp/backend/TokenAuthResult.java
```diff
@@ -0,0 +1,34 @@
...
+        this.authenticated = user.getStringID().equals(token.getUserID());
+        this.user = user;
+        this.token = token;
+        if (authenticated) {
+            this.errorMessage = null;
+        } else {
+            this.errorMessage = "Token Authentication failed";
+        }
+    }
+
```
Jacob4649, line 30: We may also want a constructor with just an error message, in case fetching the token or user from the database fails.

Jacob4649 (changes requested):
Looking really great!

backend/src/main/java/com/javahelp/backend/TokenAuthResult.java
Lam207: Good catch

Jacob4649 (approved):
Looks good!

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthManager.java
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @param userStore: interface that manages users
+     * @param tokenStore: interface that manages tokens
+     */
+    public TokenAuthManager(IUserStore userStore, ITokenStore tokenStore) {
+        this.userStore = userStore;
+        this.tokenStore = tokenStore;
+    }
+
+    /**
```
Lam207, line 24: Many of these docstrings look great! Just a few of them should be fixed (if you have time). Like this one should have `@param`

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
```diff
@@ -0,0 +1,44 @@
...
+        this.desiredUserID = desiredUserID;
+        this.authenticated = desiredUserID.equals(token.getUserID());
+        if (authenticated) {
+            this.errorMessage = null;
+        } else {
+            this.errorMessage = "Token Authentication failed";
+        }
+    }
+
+    /**
```
Lam207, line 31: Same for this one with `@param`

Enkhzul001 (dismissed):
Everything looks good!

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
BullDF: Agree, we should fix this first.

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthManager.java
BullDF: Agree!

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
```diff
@@ -0,0 +1,44 @@
...
+     * Constructor for when fetching the token from the database fails
+     */
+    public TokenAuthResult(String desiredUserID) {
+        this.errorMessage = "Could not fetch from database";
+        this.desiredUserID = desiredUserID;
+        this.token = null;
+        this.authenticated = false;
+    }
+
+    public boolean getAuthenticated() { return this.authenticated; }
```
BullDF, line 41: Is this Android Studio problem? I think we should expand this into multiple lines instead of squeezing everything in one line.

BullDF: Also, we should have Java doc for this method.

BullDF (changes requested):
Overall good but besides the `@param` issue that Lam has mentioned, I just noticed that there is one method that does not have any Java doc at all. Should add that.

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
Lam207: I think sometimes it will get automatically squeezed into one line like that, at least from my experience, and yep agree on having Javadoc, too

Lenzs: I've also thought that we shouldn't add docs to getter/setter methods, as their functions should already be very clear and readable and comments on it would only clutter up the code. There's also some discrepancy about this in the project actually, some files like ClientInfo do not have docs for getters and setters while Token does. I think we should agree on a single convention as a group and then make the necessary changes .

I also think having the entire method in one line also makes the code more readable and seems to be what was done in other classes as well.

What do you guys think?

Lenzs: For now I have just added the missing parameters and pushed the changes

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
```diff
@@ -0,0 +1,45 @@
...
+     * @param desiredUserID the ID of the user that needs to be authenticated
+     */
+    public TokenAuthResult(String desiredUserID) {
+        this.errorMessage = "Could not fetch from database";
+        this.desiredUserID = desiredUserID;
+        this.token = null;
+        this.authenticated = false;
+    }
+
+    public boolean getAuthenticated() {return this.authenticated;}
```
BullDF, line 43: Hmm Bill I think the Java doc for this method is still missing.

BullDF (changes requested):

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
Lam207: I think it does have docs for getters and setters for both `ClientUserInfo` and `Token`. Did you mean the description of private attributes? If so, the docs there is optional, and preferably not to have.

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthManager.java
```diff
@@ -0,0 +1,36 @@
...
+     * compares the input user and token and returns the result of the comparison
+     *
+     * @param desiredUserID: The ID of the user that needs to be authenticated
+     * @param token: The token being used to verify the desired user
+     *
+     * @return TokenAuthResult object with the results of the comparison
+     */
+    public TokenAuthResult authenticate(String desiredUserID, String token) {
```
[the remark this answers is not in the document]
Lam207: Um I think in the CRC cards for this, it says `String token`, and `tokenID` is actually `String id`, which seems distinct from `token` to me.

Lam207 (dismissed):
Fill out some more docstrings and it is good, but I suspect the `userinfo` package seems to have some issues.

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthResult.java
Lam207: Yeah he raised that question in the above thread

Lenzs: Yeah I addressed it in the above thread, but I've just gone and added it since that would probably be the simplest solution

BullDF: Oh yeah my bad sorry Bill I didn't notice you replied to the previous post! Thanks for adding it! Yeah for ClientInfo I think I have Java doc for the setters and getters too but I don't have doc for the attributes.

BullDF (dismissed):
Looks good! (We will probably need Jacob to remove the extra commits again since UserInfo shows up again)

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthManager.java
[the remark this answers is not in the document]
Jacob4649: > Um I think in the CRC cards for this, it says `String token`, and `tokenID` is actually `String id`, which seems distinct from `token` to me.

The current version does not have separate token ids and tokens

Jacob4649 (approved):
Looks great!

yuanshu777 (approved):
Looks good

Enkhzul001 (approved):
looks good.

PR #78, opened 2022-11-15
Login dynamo

### LINKED ISSUE #74: [Backend | Login & Auth API] Implement Login Database Methods With DynamoDB
Update current DynamoDB implementation of `IUserStore` to implement new login required methods.

Jacob4649:
Closes #74.

### DIFF
--- a/backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
+++ b/backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
@@ -14,17 +14,9 @@ import com.javahelp.model.user.UserInfo;
 import com.javahelp.model.user.UserPassword;
 import com.javahelp.model.user.UserType;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-import java.io.Serializable;
-import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
-import java.util.function.Supplier;
 
 import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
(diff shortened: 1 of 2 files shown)


backend/src/test/java/com/javahelp/backend/data/DynamoDBUserStoreTest.java
```diff
@@ -140,6 +140,46 @@ public void testPasswordUpdateRead() {
         db.delete(u.getStringID());
     }
 
+    @Test
```
Lam207, line 143: Is it OK not to have timeout time?

Lam207 (changes requested):
Hey Jacob, could you put `@Test(timeout = 50)` on the tests?

backend/src/test/java/com/javahelp/backend/data/DynamoDBUserStoreTest.java
Jacob4649: Good idea, that definitely makes sense here!

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
```diff
@@ -36,14 +30,31 @@
...
- * - userinfo: lazy Base64 serialized UserInfo
+ * - user info attributes with corresponding names
  * <p></p>
  * The structure of the items in this table should not be depended on outside of this class
  * as elements, especially the Base64 encoded serialization of {@link UserInfo} may change in the
  * future
  */
 public class DynamoDBUserStore extends DynamoDBStore implements IUserStore {
 
+    private static final String CLIENT_UPDATE = "SET username=:unameval, " +
```
Lam207, line 41: I'm not really familiar with the use of `=:`, does `SET username=:unameval, ` mean `SET username` is defined as `unameval`?

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
```diff
@@ -248,145 +315,62 @@ private static Map<String, AttributeValue> updateFromUser(User u) {
...
-
-        @Override
-        public UserInfo get() {
-            return new ClientUserInfo(email, address, phoneNumber, first, last);
-        }
+    private static boolean matchesClient(Map<String, AttributeValue> item) {
+        return item.containsKey("infotype") &&
+                "CLIENT".equals(item.get("infotype").getS()) &&
+                item.containsKey("address") &&
+                item.containsKey("firstName") &&
```
Enkhzul001, line 372: Since we are considering gender attribute, I think we should include gender here, maybe.
Also, same comment for provider info code above line 355.

Enkhzul001 (changes requested):

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
Jacob4649: This is just the syntax for DynamoDB update item requests. `=:` is not an operator in this context (note that in some contexts `:=` is used as an assignment operator), rather the field named `username` is being assigned the value called `:unameval`.

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
Jacob4649: I will add these once #80 is merged!

BullDF (approved):
Good stuff!

Enkhzul001 (approved):
Great work

Lenzs (approved):
Looks really good!

backend/src/test/java/com/javahelp/backend/data/DynamoDBUserStoreTest.java
```diff
@@ -25,7 +25,7 @@ public class DynamoDBUserStoreTest {
 
     DynamoDBUserStore db = new DynamoDBUserStore(tableName, regions);
 
-    @Test
+    @Test(timeout = 5000)
```
Lam207, line 28: Sorry for reviewing late. Just a little question, large timeout limit like this is to make sure that everything won't crash (even when the execution speed is slow?)

Lam207 (approved):
Looks great!

PR #80, opened 2022-11-16
Addition of `gender` attribute 

Lam207:
Add a new `gender` attribute to `ProviderInfo` to be used as a filter feature later.

### DIFF
--- a/model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
+++ b/model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
@@ -8,6 +8,7 @@ public class ProviderUserInfo extends UserInfo {
     private boolean certified;
     private String address;
     private String phoneNumber;
+    private String gender;
 
     /**
      * Constructs a ProviderInfo object.
@@ -18,12 +19,18 @@ public class ProviderUserInfo extends UserInfo {
      * @param practiceName: the name of the provider.
      */
     public ProviderUserInfo(String emailAddress, String address, String phoneNumber,
-                            String practiceName) {
+                            String practiceName, String gender) {
         super(emailAddress);
         this.practiceName = practiceName;
         this.certified = false;
         this.address = address;
         this.phoneNumber = phoneNumber;
+        if (!(gender.toUpperCase().equals("MALE") || gender.toUpperCase().equals("FEMALE") ||
+                gender.toUpperCase().equals("OTHERS"))){
+            throw new IllegalArgumentException("This is not a valid input. Please try \"MALE\"" +
+                    "or \"FEMALE\" or \"OTHERS\"");
+        }
+        this.gender = gender.toUpperCase();
     }
 
     /**
@@ -96,6 +103,24 @@ public class ProviderUserInfo extends UserInfo {
      */
     public void setPhoneNumber(String phoneNumber) {
         this.phoneNumber = phoneNumber;
+    }
+
+    /**
+     * Gets the gender of this provider
+     *
+     * @return the current gender of this client
+     */
+    public String getGender(){
+        return gender;
+    }
+
+    /**
+     * Sets a new gender to this client
+     *
+     * @param gender an updated gender of this client
+     */
+    public void setGender(String gender){
+        this.gender = gender.toUpperCase();
     }
 
     @Override
--- a/model/src/test/java/com/javahelp/model/user/ProviderUserInfoTest.java
+++ b/model/src/test/java/com/javahelp/model/user/ProviderUserInfoTest.java
@@ -13,7 +13,7 @@ public class ProviderUserInfoTest {
     @Before
     public void setUp() {
         p = new ProviderUserInfo("uoft@utoronto.ca", "St. George",
-                "111-234-5678", "University of Toronto");
+                "111-234-5678", "University of Toronto", "male");
     }
 
     @Test(timeout = 50)
@@ -49,6 +49,12 @@ public class ProviderUserInfoTest {
         assertFalse(p.isCertified());
         p.setCertified(true);
         assertTrue(p.isCertified());
+    }
+
+    @Test(timeout = 50)
+    public void testSetGender(){
+        p.setGender("female");
+        assertEquals("FEMALE", p.getGender());
     }
 
     @Test(timeout = 50)


model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -98,6 +100,30 @@ public void setPhoneNumber(String phoneNumber) {
...
+     */
+    public String setGender(String gender){
+        if (!(gender.toUpperCase().equals("MALE") || gender.toUpperCase().equals("FEMALE") ||
+                gender.toUpperCase().equals("OTHERS"))){
+            throw new IllegalArgumentException("This is not a valid input. Please try \"MALE\"" +
+                    "or \"FEMALE\" or \"OTHERS\"");
+        }
+        this.gender = gender.toUpperCase();
+        return null;
+    }
```
Jacob4649, line 125: Instead of a `String` it might be a good idea to either make this field some kind of enum type with `MALE`, `FEMALE`, and `OTHER` fields, or to make it a `String` field without restriction. I don't think we should make it a `String` and restrict the possible values.

model/src/test/java/com/javahelp/model/user/ProviderUserInfoTest.java
```diff
@@ -51,6 +51,19 @@ public void testSetCertified() {
...
+    @Test(timeout = 50)
+    public void testSetGender(){
+        p.setGender("female");
+        assertEquals("FEMALE", p.getGender());
+
+        try{
+            p.setGender("ABC");
+        }
+        catch(IllegalArgumentException e){
+            System.out.println("Invalid input for gender. Please try again");
```
Jacob4649, line 63: When testing for exceptions, instead of using try catches, which do not confirm an exception was actually caught, try using the `expected` field in the `@Test` annotation.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
Lam207: Sure let me do it

model/src/test/java/com/javahelp/model/user/ProviderUserInfoTest.java
Lam207: Sure let me apply that

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -98,6 +100,30 @@ public void setPhoneNumber(String phoneNumber) {
...
+     * @param gender an updated gender of this client
+     */
+    public void setGender(String gender){
+        if (!(Gender.FEMALE.toString().equals(gender.toUpperCase()) ||
+                Gender.MALE.toString().equals(gender.toUpperCase()) ||
+                Gender.OTHERS.toString().equals(gender.toUpperCase()))){
+            throw new IllegalArgumentException("This is not a valid input. Please try \"MALE\"" +
+                    "or \"FEMALE\" or \"OTHERS\"");
+        }
+        this.gender = gender.toUpperCase();
```
Jacob4649, line 124: Instead of setting and storing gender as a `String`, if you are going to have an enum for it, it probably makes more sense to just set/store it as an enum. So instead of `String gender`, you would have `Gender gender`.

Jacob4649 (changes requested):

model/src/test/java/com/javahelp/model/user/ProviderUserInfoTest.java
```diff
@@ -51,6 +53,17 @@ public void testSetCertified() {
         assertTrue(p.isCertified());
     }
 
+    @Test(timeout = 50)
+    public void testSetGender(){
+        p.setGender("female");
```
BullDF, line 58: Here too.

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
```diff
@@ -351,7 +351,8 @@ public SerializableProviderUserInfo(ProviderUserInfo base) {
 
         @Override
         public UserInfo get() {
-            ProviderUserInfo p = new ProviderUserInfo(email, address, phoneNumber, practice);
+            ProviderUserInfo p = new ProviderUserInfo(email, address, phoneNumber, practice,
+                    "");
```
BullDF, line 355: Probably here too.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
BullDF, line 124: This does not make sense if using `enum`.

BullDF (changes requested):
Good looking code Lam! I have added specific places that you have to change the type `String` to type `Gender`, as per Jacob's comments.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -8,6 +8,7 @@ public class ProviderUserInfo extends UserInfo {
     private boolean certified;
     private String address;
     private String phoneNumber;
+    private String gender;
```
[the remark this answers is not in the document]
Lam207: Thanks for your comment too. Sorry I was busy today so didn't make change quickly

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
Lam207: Gotcha

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -98,6 +113,29 @@ public void setPhoneNumber(String phoneNumber) {
...
+     * Sets a new gender to this client
+     *
+     * @param gender an updated gender of this client
+     */
+    public void setGender(Gender gender){
+        if (!(Gender.FEMALE.equals(gender) || Gender.MALE.equals(gender) ||
+                Gender.OTHERS.equals(gender))){
+            throw new IllegalArgumentException("This is not a valid input. Please try \"MALE\"" +
+                    "or \"FEMALE\" or \"OTHERS\"");
+        }
```
Jacob4649, line 135: This is no longer necessary since the input is an enum, so it will always be one of those three values, or null. In either case, I don't think this check is still necessary.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -1,5 +1,7 @@
 package com.javahelp.model.user;
 
+import java.security.Provider;
```
Jacob4649, line 3: I don't think this import is needed anywhere

Jacob4649 (changes requested):
Looks great!

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -16,14 +17,26 @@ public class ProviderUserInfo extends UserInfo {
...
+                            String practiceName, Gender gender) {
         super(emailAddress);
         this.practiceName = practiceName;
         this.certified = false;
         this.address = address;
         this.phoneNumber = phoneNumber;
+        this.setGender(gender);
+    }
+
+    public ProviderUserInfo(String emailAddress, String address, String phoneNumber,
```
Jacob4649, line 32: Make sure to document this constructor!

Jacob4649 (changes requested):

Jacob4649 (dismissed):
Looks good!

model/src/main/java/com/javahelp/model/user/Gender.java
```diff
@@ -0,0 +1,7 @@
+package com.javahelp.model.user;
+
+public enum Gender {
+    MALE,
+    FEMALE,
+    OTHERS
```
Lenzs, line 6: just a thought, but would it be possible to make it so that users can specify their gender if they pick other?

Lam207: I know there are a bunch of other groups too, but how do you all think? Is `OTHERS` needed to be more specific?

Jacob4649: > just a thought, but would it be possible to make it so that users can specify their gender if they pick other?

Ye that's a really great argument for making it an unrestricted string

Enkhzul001 (dismissed):
Looks great! And good idea that you considered including gender attribtue.

model/src/main/java/com/javahelp/model/user/Gender.java
yuanshu777: Maybe, we can do something like male, female, and if other, please specify...
In the UI, it would be buttons and if they pick other, then a text would jump up and they can type their genders

yuanshu777 (changes requested):
We might need some restriction for the string that the user put for their gender if we want to have more specific genders rather than other

model/src/main/java/com/javahelp/model/user/Gender.java
Lam207: I discussed with Professor Jonathan, and he said when `OTHERS` is selected, there will be another option for them to enter a costume string that represents their own gender.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -8,22 +8,45 @@ public class ProviderUserInfo extends UserInfo {
     private boolean certified;
     private String address;
     private String phoneNumber;
+    private Gender gender;
```
BullDF, line 11: Just a thought, since we have the attribute as `practiceName` for providers, do we allow a company/organization of therapists to register one account? If yes then it does not make sense to have a gender for this company. But if no, I think we should change `practiceName` into `firstName` and `lastName` just like the client.

BullDF (commented):
Just some thoughts on the gender attribute.

model/src/main/java/com/javahelp/model/user/ClientUserInfo.java
```diff
@@ -12,11 +12,11 @@ public class ClientUserInfo extends UserInfo {
...
-     * @param emailAddress: the email address of this client.
-     * @param address: the address of this client.
-     * @param phoneNumber: the phone number of this client.
-     * @param firstName: the first name of this client.
-     * @param lastName: the last name of this client.
+     * @param emailAddress the email address of this client.
+     * @param address the address of this client.
+     * @param phoneNumber the phone number of this client.
+     * @param firstName the first name of this client.
+     * @param lastName the last name of this client.
```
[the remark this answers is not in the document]
Lam207: Yeah it is also a topic that I discussed with Owen earlier in the group chat. You should look at that for two thoughts. What do you all think? Have `Gender` attribute to Clients, too, or like now?

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
Lam207: Wait, I thought I designed to assign `Gender` as an attribute to a Provider User, not a company?

model/src/main/java/com/javahelp/model/user/ClientUserInfo.java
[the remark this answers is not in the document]
Lam207: I honestly think this is only necessary to clients, as `Clients` will `search` and `filter` based on the info (given by `Providers`) for providers that fit their inserted info. In other words, only `Providers` will need to fill out `Gender`, for `Clients`, it is kind of pointless.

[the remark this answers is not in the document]
stevolopolis: I agree that `Gender` should also be added to Clients

stevolopolis (approved):
Generally an approvable PR. But will need to add `Gender` to Client soon in another PR.

Lenzs (dismissed):
Everything looks great!

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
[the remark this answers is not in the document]
Lam207: Oh it was my mistake. I should have converted it to `Gender` enum, and used it as `Gender.GENDER`

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
```diff
@@ -334,14 +338,17 @@ private static User userFromDynamo(Map<String, AttributeValue> item) {
                     firstName = item.get("firstName").getS(),
                     lastName = item.get("lastName").getS(),
                     address = item.get("address").getS();
-            info = new ClientUserInfo(email, address, phoneNumber, firstName, lastName);
+            info = new ClientUserInfo(email, address, phoneNumber, firstName, lastName,
+                    Gender.FEMALE);
```
yuanshu777, line 342: I'm not too sure why you specify Gender with FEMALE at here.

model/src/main/java/com/javahelp/model/user/ClientUserInfo.java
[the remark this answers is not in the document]
yuanshu777: Yes, even **Providers** do not need the function of **search** and **filter**, It's still good for them to choose a gender. Not only in the social aspect, but also that we might need the client's gender for the survey, or for the purpose of the providers providing better counseling services.

model/src/main/java/com/javahelp/model/user/Gender.java
```diff
@@ -0,0 +1,8 @@
+package com.javahelp.model.user;
+
+public enum Gender {
+    MALE,
+    FEMALE,
+    OTHERS,
+    UNMENTIONED
```
yuanshu777, line 7: The unmentioned option will be corresponding to the "prefer not to tell" button, right?

yuanshu777 (approved):
Thanks for adding the gender attribute

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
Lam207: To fill out the `Gender` parameter for this `Client` :)

model/src/main/java/com/javahelp/model/user/Gender.java
Lam207: Yup, and it can be described default value.

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
Jacob4649: I also don't understand why you specify female, why not get the gender attribute from the database, parse it to the gender enum, and assign it to the user info?

Jacob4649 (changes requested):

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
Lam207: I see

Enkhzul001 (approved):
looks good

Lam207:
This branch is permanently removed because now users could describe and fill out their `gender` status, if wanted, on `Survey` answers instead. That approach is better and more flexible than this one, because it does not require going back to the `UserInfo` to add this additional feature.

PR #82, opened 2022-11-17
Add the client registration activity, edit the style of UI.

Enkhzul001:
I added client registration activity, and changed some of the style. I also added back arrow.

### DIFF
--- a/app/src/main/java/com/javahelp/frontend/MainActivity.java
+++ b/app/src/main/java/com/javahelp/frontend/MainActivity.java
@@ -28,7 +28,7 @@ public class MainActivity extends AppCompatActivity {
         //find the provider register button by its id
         buttonpr = (Button) findViewById(R.id.bot_p_reg);
 
-        // link the button to the second acticity(front page)
+        // link the button to the second activity(front page)
         buttonLogin.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
@@ -38,7 +38,7 @@ public class MainActivity extends AppCompatActivity {
             }
         });
 
-        // link the button to the second acticity(front page)
+        // link the button to the second activity(front page)
         buttonpr.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/javahelp/frontend/MainActivity.java
```diff
@@ -48,7 +48,15 @@ public void onClick(View view) {
             }
         });
 
-
+        // link the button to the second activity(front page)
+        buttoncr.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                Intent intent = null;
```
Lam207, line 55: Could these two be combined into one step?

Lam207: The one above and below this message. Sorry this splits them apart. 
Like  `Intent intent = new Intent(MainActivity.this,ClientRegistrationActivity.class);`. Is there any specific reason behind separating these two?

Enkhzul001: I am bit confused. Can you explain more?

Enkhzul001: Ah wait, I see what you mean. I can combine them.

yuanshu777: I think they mean the same, and I think both ways are ok

yuanshu777 (approved):
It looks good to me.
Thanks for adjust the style

Lam207 (approved):
Looks great!

BullDF (commented):
I am not really familiar with the UI part so I am not sure where the specific codes are but I have three things in mind now:

1. Should we add a Verify Password field for the user to enter their password again?
2. Is it possible to set the `Gender` field to be a dropdown menu or multiple-choice in align with the `Gender` class we just added? We can have the text entry box for the "other gender" option if we decided to have that freedom. I just think that it will be very difficult to record that into the database if the user can enter whatever they want.
3. Probably rearrange the entry boxes? Would it make sense to enter the email and password first before personal information? I think this is what the majority of website do.

Thanks for the work Zula!

Lenzs (approved):
Everything looks great!

PR #84, opened 2022-11-18
Filter and search v1 (attributeMaps, QueryConstraints on DynamoDB, Constraint Classes)

stevolopolis:
Changes:

1. Added attributeMap to userInfo
2. Added and amended methods in DynamoDBUserStore to create, update, and read attributeMaps
3. Amended DynamoDBUserStoreTest, ProviderUserInfoTest, ClientUserInfoTest to accommodate the addition of attributeMaps
4. Added method in DynamoDBUSerStore to query using keys in attributeMaps
5. Created interface, abtract class for Constraints. Also added some specialty Constraint classes which could be useful depending on front-end implementations.
6. Added and passed UserQueryConstraintTest which tests out all Constraint classes


backend/src/test/java/com/javahelp/backend/data/DynamoDBUserStoreTest.java
```diff
@@ -34,6 +38,8 @@ public void testCreateRead() {
                 "289375034875093",
                 "Erin",
                 "McDonald");
+        clientInfo.setAttribute("Age", "35");
+        clientInfo.setAttribute("Specialty", "Pain");
         User u = new User("test", clientInfo, "test_user");
```
Lam207, line 43: This `StringID` should be randomly generated by the User database, but this is fine with me.

model/src/main/java/com/javahelp/model/user/UserInfo.java
```diff
@@ -33,6 +36,33 @@ public void setEmailAddress(String emailAddress) {
         this.emailAddress = emailAddress;
     }
 
+    /**
```
Lam207, line 39: Should these be moved to `ClientUserInfo` instead?

model/src/test/java/com/javahelp/model/user/ClientUserInfoTest.java
```diff
@@ -51,6 +53,32 @@ public void testSetAddress() {
         assertEquals("Bahen Center", c.getAddress());
     }
 
+    @Test(timeout = 50)
+    public void testSetSingleAttributeExist() {
+        c.setAttribute("Gender", "Male");
```
Lam207, line 58: With `Gender` pending PR, I think this would also need to change soon after that PR is approved, but it is still fine right now.

Lam207 (changes requested):

backend/src/test/java/com/javahelp/backend/data/DynamoDBUserStoreTest.java
stevolopolis: Correct me if I'm wrong, but I believe it is indeed generated by the database in this case.

model/src/main/java/com/javahelp/model/user/UserInfo.java
stevolopolis: ProviderUserInfo also needs the setEmailAddress method. If you look at the constructor for ProviderUserInfo, emailAddress is added to it as a private variable too.

backend/src/main/java/com/javahelp/backend/query/VanillaConstraint.java
```diff
@@ -0,0 +1,10 @@
+package com.javahelp.backend.query;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class VanillaConstraint extends Constraint{
```
Enkhzul001, line 6: What is a VanillaConstraint?

backend/src/test/java/com/javahelp/backend/query/UserQueryConstraintTest.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Random;
+import java.util.Set;
+import java.util.HashSet;
+
+public class UserQueryConstraintTest {
+    String tableName = "javahelpBackendUsers";
+    Regions regions = Regions.US_EAST_1;
+    DynamoDBUserStore db = new DynamoDBUserStore(tableName, regions);
+
+    @Test
```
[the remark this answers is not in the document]
Enkhzul001: like timeout = 50

Enkhzul001 (approved):
Just left a few comments.
Overall, great work!

model/src/main/java/com/javahelp/model/user/UserInfo.java
Lam207: Oh the position of my message is displayed is a bit misunderstanding. I meant the one below, since both `searching` and `matching` are from Client UI.

backend/src/main/java/com/javahelp/backend/query/VanillaConstraint.java
stevolopolis: It is just a very simply constraint that finds matching attribute values through equivalence relation.

model/src/main/java/com/javahelp/model/user/UserInfo.java
stevolopolis: We'll have to get the attributeMaps of both Client and Providers to calculate similarities. I believe this is necessary for the ProviderRanker method.

model/src/test/java/com/javahelp/model/user/ClientUserInfoTest.java
stevolopolis: Agreed

model/src/main/java/com/javahelp/model/user/UserInfo.java
Lam207: Alright gotcha

stevolopolis: If there are not other issues from your end, do remember to approve this PR

backend/src/test/java/com/javahelp/backend/query/UserQueryConstraintTest.java
```diff
@@ -0,0 +1,229 @@
...
+            actualProvidersSet.add(user.getStringID());
+        }
+
+        assertEquals(expectedProvidersSet, actualProvidersSet);
+
+        db.delete(u1.getStringID());
+        db.delete(u2.getStringID());
+    }
+
+    @Test(timeout = 5000)
```
Enkhzul001, line 70: Is the timeout suppose to be this long?

Enkhzul001 (approved):
Great work

backend/src/test/java/com/javahelp/backend/query/UserQueryConstraintTest.java
Lam207: I guess, to prevent the worst cases when there're connecting issues to Gradle or something else.

backend/build.gradle
```diff
@@ -20,9 +20,9 @@ dependencies {
     implementation 'com.amazonaws:aws-java-sdk-dynamodb'
 
     api project(path: ':model')
+    implementation 'com.google.firebase:firebase-database:20.0.4'
```
Jacob4649, line 23: I'm not sure I see the need for a second firebase database.

Jacob4649 (changes requested):

backend/src/test/java/com/javahelp/backend/query/UserQueryConstraintTest.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.Set;
+import java.util.HashSet;
+
+public class UserQueryConstraintTest {
+    String tableName = "javahelpBackendUsers";
+    Regions regions = Regions.US_EAST_1;
+    DynamoDBUserStore db = new DynamoDBUserStore(tableName, regions);
```
Jacob4649, line 25: Should instantiate through `getDefaultInstance` instead

backend/src/main/java/com/javahelp/backend/data/DynamoDBUserStore.java
```diff
@@ -63,7 +70,7 @@ public class DynamoDBUserStore extends DynamoDBStore implements IUserStore {
      * @param tableName {@link String} table name to be used
      * @param region    {@link Regions} to be used
      */
-    DynamoDBUserStore(String tableName, Regions region) {
+    public DynamoDBUserStore(String tableName, Regions region) {
```
Jacob4649, line 73: We may want to keep the implementation private

backend/src/main/java/com/javahelp/backend/query/IConstraint.java
```diff
@@ -0,0 +1,11 @@
+package com.javahelp.backend.query;
+
+import java.util.ArrayList;
+import java.util.Map;
+
+public interface IConstraint {
```
Jacob4649, line 6: Does this interface add value beyond the abstract class? Will there ever be a `IConstraint` that does not extend the abstract class?

backend/src/main/java/com/javahelp/backend/query/VanillaConstraint.java
Jacob4649, line 6: Perhaps a better naming scheme would be `ConstraintBase` for the abstract class, and `Constraint` for this class

backend/src/main/java/com/javahelp/backend/query/UserQueryConstraint.java
```diff
@@ -0,0 +1,75 @@
...
+
+            HashSet<User> subsetProviders = getProvidersWithConstraint(constraint);
+            HashSet<String> subsetProvidersId = IdSet(subsetProviders);
+            if (providers.isEmpty()) {
+                providers = subsetProviders;
+            }
+            else {
+                providersId.retainAll(subsetProvidersId);
+                providers = getUserSetFromIdSet(providers, providersId);
+            }
```
Jacob4649, line 41: Instead of running multiple queries and combining the results, it is probably better to run a single query.

model/src/main/java/com/javahelp/model/user/ClientUserInfo.java
```diff
@@ -8,6 +10,7 @@ public class ClientUserInfo extends UserInfo {
     private String lastName;
     private String address;
     private String phoneNumber;
+    private HashMap<String, Integer> attributeMap;
```
Jacob4649, line 13: Rather than a map of attributes, we probably want these to be fields in the class

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -8,6 +10,7 @@ public class ProviderUserInfo extends UserInfo {
     private boolean certified;
     private String address;
     private String phoneNumber;
+    private HashMap<String, Integer> attributeMap;
```
Jacob4649, line 13: Same as with `ClientUserInfo`, any information that goes here should probably be a field in the class.

Jacob4649 (changes requested):
You probably also want constraint classes to be independent of the database. Right now they are very tied to DynamoDB, in some places even unnecessarily.

model/src/test/java/com/javahelp/model/user/ClientUserInfoTest.java
```diff
@@ -51,6 +53,32 @@ public void testSetAddress() {
...
+
+    @Test(timeout = 50)
+    public void testSetSingleAttributeNonExist() {
+        c.setAttribute("Gender", "Male");
+        assertEquals("-1", c.getSingleAttribute("Age"));
+    }
+
+    @Test(timeout = 50)
+    public void testSetMultipleAttribute() {
+        c.setAttribute("Gender", "Male");
```
Lam207, line 70: If `Gender` enum is approved, then this line should be Gender.MALE, am I right? Just in case when this is merged before `Gender` PR.

backend/build.gradle
stevolopolis: Agreed and will be changed.

backend/src/main/java/com/javahelp/backend/query/IConstraint.java
stevolopolis: Agreed. Could be deleted.

backend/src/main/java/com/javahelp/backend/query/UserQueryConstraint.java
stevolopolis: Changed in recent push

PR #85, opened 2022-11-19
Client UI activity

Enkhzul001:
I added password match, arranged the order of the forms, and made gender have dropdown selection feature.

### DIFF
--- a/app/src/main/java/com/javahelp/frontend/ClientRegistrationActivity.java
+++ b/app/src/main/java/com/javahelp/frontend/ClientRegistrationActivity.java
@@ -1,28 +1,20 @@
 package com.javahelp.frontend;
 
-import androidx.annotation.ColorInt;
 import androidx.appcompat.app.AppCompatActivity;
-
 import android.annotation.SuppressLint;
 import android.content.Intent;
-import android.graphics.Color;
 import android.os.Bundle;
 import android.view.View;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
-import android.widget.EditText;
 import android.widget.Spinner;
-import android.widget.TextView;
 
 import com.javahelp.R;
 
-import java.util.ArrayList;
 
 
 public class ClientRegistrationActivity extends AppCompatActivity {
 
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/javahelp/frontend/ClientRegistrationActivity.java
```diff
@@ -1,15 +1,133 @@
...
+    TextView passwordAlert, confirmAlert, emailAlert, userAlert, firstAlert, lastAlert, homeAlert, phoneAlert;
+    private Button buttonsignup;
+
+    @SuppressLint("MissingInflatedId")
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_client_registration);
+        //find password, confirm password, name, email id, home address, phone, signup button
+        user = findViewById(R.id.username);
```
yuanshu777, line 34: for this, we can use data binding to make our code cleaner

yuanshu777: I'm doing this

yuanshu777 (approved):
Everything looks great. It's just that we can use data binding to make the code easier to track. I'M working on this right now

app/src/main/java/com/javahelp/frontend/ClientRegistrationActivity.java
Enkhzul001: Oh I see. Yeah that would make things cleaner. Let me just make a new commit.

app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,9 @@
 <resources>
     <string name="app_name">JH Mental Helper</string>
+    <string-array name="genders">
+        <item>Gender</item>
+        <item>Female</item>
```
Lam207, line 5: I guess you could add `Unmentioned` field here, too, after my PR is approved. Yeah sorry, the updated version of `Gender` is still pending on request.

PR #96, opened 2022-11-20
Token abstract handler

### LINKED ISSUE #94: [Backend | Login & Auth API] Add Token Handler Abstract Class In Backend

Jacob4649:
Closes #94.


Enkhzul001 (dismissed):
looks good.

model/src/main/java/com/javahelp/model/token/Token.java
```diff
@@ -56,9 +57,10 @@ public Token(String token, Instant issued, Instant expiry, String tag, String us
     /**
      * Creates a new {@link Token} for the given user, with a random token string,
      * initiated at the current moment.
+     *
      * @param validDuration {@link Duration} the {@link Token} is valid for
-     * @param tag {@link String} tags for the {@link Token}
-     * @param userId {@link String} user id for the {@link Token}
+     * @param tag           {@link String} tags for the {@link Token}
```
Lam207, line 62: What happened with these two blocks though?

stevolopolis (approved):
Code looks good to me. Only concern would be should we add unit tests for failed authentications (e.g. no user)?

backend/src/main/java/com/javahelp/backend/domain/user/authentication/TokenAuthManager.java
```diff
@@ -22,12 +25,21 @@ public TokenAuthManager(IUserStore userStore, ITokenStore tokenStore) {
...
+     * @param input {@link ITokenAuthInput} supplying input information
      * @return TokenAuthResult object with the results of the comparison
      */
-    public TokenAuthResult authenticate(String desiredUserID, String token) {
-        this.authenticate = new TokenAuthResult(desiredUserID, tokenStore.read(token));
+    public TokenAuthResult authenticate(ITokenAuthInput input) {
+        User u = userStore.read(input.getUserID());
+        Token t = tokenStore.read(input.getToken());
+        String l = input.getToken();
+        if (u == null) {
```
Lam207, line 36: Looks great!

Lam207 (dismissed):
Looks good to me

model/src/main/java/com/javahelp/model/token/Token.java
Jacob4649: its just auto formatting

Jacob4649:
> Code looks good to me. Only concern would be should we add unit tests for failed authentications (e.g. no user)?

I want to, but its probably best to wait for an implementation before adding these, just so that we don't need to write a whole new test handler

Enkhzul001 (approved):
looks good

Lam207 (approved):
Great work!

backend/src/main/java/com/javahelp/backend/endpoint/APIGatewayResponse.java
```diff
@@ -18,10 +18,18 @@
...
+     * Response code for unauthenticated requests
+     */
+    public static int UNAUTHENTICATED = 401;
+
     /**
      * Response code for unsupported media type errors
-     * <p>
-     * Should be returned by NonJSONBodyResponse
      */
     public static int UNSUPPORTED_MEDIA_TYPE = 415;
```
BullDF, line 34: Should we set these variables to final since the status code are set and constant?

BullDF (approved):
Only concern is the final status of the status code.

PR #98, opened 2022-11-20
User register for both Clients and Providers

Lam207:
A backend register user use case for all types of `users`, such as for `clients` and `providers`

### DIFF
--- a/backend/src/main/java/com/javahelp/backend/domain/user/register/IUserRegisterInputBoundary.java
+++ b/backend/src/main/java/com/javahelp/backend/domain/user/register/IUserRegisterInputBoundary.java
@@ -1,0 +1,26 @@
+package com.javahelp.backend.domain.user.register;
+
+import com.javahelp.model.user.User;
+
+/**
+ * Information necessary to register a {@link User}
+ */
+public interface IUserRegisterInputBoundary {
+
+    /**
+     * @return the {@link User}'s email address
+     */
+    String getEmailAddress();
+
+    /**
+     * @return the {@link User}'s username
+     */
+    String getUsername();
+
+    /**
+     * @return base64 encoded byte array with the first 4 bytes containing a 32 bit integer specifying
+     * salt length, the next bytes up to the previously specified length of the salt specifying the salt,
+     * and the last bytes specifying the hash
+     */
+    String getSaltAndHash();
+}
--- a/backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
+++ b/backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
@@ -1,0 +1,63 @@
+package com.javahelp.backend.domain.user.register;
+
+import com.javahelp.model.user.User;
+import com.javahelp.model.user.UserPassword;
+
+/**
+ * Response for {@link User} registration from a {@link UserRegisterInteractor}
+ */
+public class RegisterUserResponse {
+
+    private User user = null;
+    private UserPassword password = null;
+    private String error = null;
+
+    /**
+     * For bundle up info of the result of this successful registration
+     * @param user the newly created {@link User}
+     * @param password the newly created {@link UserPassword}
+     */
+    public RegisterUserResponse(User user, UserPassword password) {
+        this.user = user;
+        this.password = password;
+    }
+
+    /**
+     * An overloaded constructor to make a report for this failed registration
+     * @param error a description of failure at registration
+     */
+    public RegisterUserResponse(String error) {
+        this.error = error;
+    }
+
+    /**
+     * @return a description of failure at registering from {@link UserRegisterInteractor}.
+     * Null if succeeded
+     */
+    public String getErrorMessage() {
+        return error;
+    }
+
+    /**
+     * @return an instance of a newly created {@link User}
+     * Null if failed
+     */
+    public User getUser() {
+        return user;
+    }
+
+    /**
+     * @return an instance of a newly made {@link UserPassword}
+     * Null if failed
+     */
+    public UserPassword getPassword() {
+        return password;
+    }
+
+    /**
+     * @return whether the user succeeded at registering or not
+     */
+    public boolean isSuccess() {
+        return error == null;
+    }
+}
--- a/backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
+++ b/backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
@@ -28,17 +28,17 @@ public abstract class UserRegisterInteractor <T extends IUserRegisterInputBounda
      * Checks whether email and username are taken
      * @param boundary {@link IUserRegisterInputBoundary} with input
      *
-     * @return a {@link RegisterUserResponse} with error message, or else null
+     * @return a {@link UserRegisterResult} with error message, or else null
      */
-    private RegisterUserResponse areEmailAndUsernameTaken(T boundary) {
+    private UserRegisterResult areEmailAndUsernameTaken(T boundary) {
         User emailTaken = userStore.readByEmail(boundary.getEmailAddress());
         User usernameTaken = userStore.readByUsername(boundary.getUsername());
         if (emailTaken != null && usernameTaken != null){
-            return new RegisterUserResponse("Email and username already taken");
+            return new UserRegisterResult("Email and username already taken");
         } else if (emailTaken != null){
-            return new RegisterUserResponse("Email already taken");
+            return new UserRegisterResult("Email already taken");
         } else if (usernameTaken != null){
-            return new RegisterUserResponse("Username already taken");
+            return new UserRegisterResult("Username already taken");
         }
 
         // none already taken
@@ -49,10 +49,10 @@ public abstract class UserRegisterInteractor <T extends IUserRegisterInputBounda
      * Determines whether the provided input can assemble a valid {@link UserInfo}
      * @param boundary <T> {@link IUserRegisterInputBoundary} to retrieve input
      *
-     * @return a {@link RegisterUserResponse} containing the error message
+     * @return a {@link UserRegisterResult} containing the error message
      * for invalid {@link IUserRegisterInputBoundary}, or else null
      */
-    protected abstract RegisterUserResponse isValidated(T boundary);
+    protected abstract UserRegisterResult isValidated(T boundary);
 
     /**
      * Assembles {@link UserInfo} from input
@@ -65,32 +65,32 @@ public abstract class UserRegisterInteractor <T extends IUserRegisterInputBounda
      * Creates and registers a {@link User}
      * @param boundary <T> valid {@link IUserRegisterInputBoundary}
      *
-     * @return a {@link RegisterUserResponse} with the registration results
+     * @return a {@link UserRegisterResult} with the registration results
      */
-    private RegisterUserResponse registerUserToDatabase(T boundary) {
+    private UserRegisterResult registerUserToDatabase(T boundary) {
         UserInfo info = bundleUserInfo(boundary);
         User user = new User("", info, boundary.getUsername());
 
         UserPassword userPassword = new UserPassword(boundary.getSaltAndHash());
         User newUser = userStore.create(user, userPassword);
 
-        return new RegisterUserResponse(newUser, userPassword);
+        return new UserRegisterResult(newUser, userPassword);
     }
 
     /**
      * Registers a new {@link User}
      * @param boundary <T> {@link IUserRegisterInputBoundary} providing input
-     * @return a new {@link RegisterUserResponse} with created {@link User}, or an error message
+     * @return a new {@link UserRegisterResult} with created {@link User}, or an error message
      */
-    public RegisterUserResponse register(T boundary) {
+    public UserRegisterResult register(T boundary) {
         // check email and username are not taken
-        RegisterUserResponse emailUsernameTaken = areEmailAndUsernameTaken(boundary);
+        UserRegisterResult emailUsernameTaken = areEmailAndUsernameTaken(boundary);
         if (emailUsernameTaken != null) {
             return emailUsernameTaken;
         }
 
         // Check user input valid
-        RegisterUserResponse inputInvalid = isValidated(boundary);
+        UserRegisterResult inputInvalid = isValidated(boundary);
         if (inputInvalid != null) {
             return inputInvalid;
         }
--- a/backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
+++ b/backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
@@ -1,0 +1,101 @@
+package com.javahelp.backend.domain.user.register;
+
+import com.javahelp.backend.data.IUserStore;
+import com.javahelp.model.user.User;
+import com.javahelp.model.user.UserInfo;
+import com.javahelp.model.user.UserPassword;
+
+/**
+ * An abstract interactor to make new {@link User} and {@link UserPassword} a specific type of {@link User}
+ * @param <T> an interface supplying registration info for a {@link User}
+ * @param <U> type of {@link UserInfo} for {@link User} type to register
+ */
+public abstract class UserRegisterInteractor <T extends IUserRegisterInputBoundary,
+        U extends UserInfo>{
+
+    private final IUserStore userStore;
+
+    /**
+     * Creates a new {@link UserRegisterInteractor} instance
+     *
+     * @param userStore  {@link IUserStore} to use
+     */
+    public UserRegisterInteractor(IUserStore userStore) {
+        this.userStore = userStore;
+    }
+
+    /**
+     * Checks whether email and username are taken
+     * @param boundary {@link IUserRegisterInputBoundary} with input
+     *
+     * @return a {@link RegisterUserResponse} with error message, or else null
+     */
+    private RegisterUserResponse areEmailAndUsernameTaken(T boundary) {
+        User emailTaken = userStore.readByEmail(boundary.getEmailAddress());
+        User usernameTaken = userStore.readByUsername(boundary.getUsername());
+        if (emailTaken != null && usernameTaken != null){
+            return new RegisterUserResponse("Email and username already taken");
+        } else if (emailTaken != null){
+            return new RegisterUserResponse("Email already taken");
+        } else if (usernameTaken != null){
+            return new RegisterUserResponse("Username already taken");
+        }
+
+        // none already taken
+        return null;
+    }
+
+    /**
+     * Determines whether the provided input can assemble a valid {@link UserInfo}
+     * @param boundary <T> {@link IUserRegisterInputBoundary} to retrieve input
+     *
+     * @return a {@link RegisterUserResponse} containing the error message
+     * for invalid {@link IUserRegisterInputBoundary}, or else null
+     */
+    protected abstract RegisterUserResponse isValidated(T boundary);
+
+    /**
+     * Assembles {@link UserInfo} from input
+     *
+     * @return a {@link IUserRegisterInputBoundary} with input data
+     */
+    protected abstract U bundleUserInfo(T boundary);
+
+    /**
+     * Creates and registers a {@link User}
+     * @param boundary <T> valid {@link IUserRegisterInputBoundary}
+     *
+     * @return a {@link RegisterUserResponse} with the registration results
+     */
+    private RegisterUserResponse registerUserToDatabase(T boundary) {
+        UserInfo info = bundleUserInfo(boundary);
+        User user = new User("", info, boundary.getUsername());
+
+        UserPassword userPassword = new UserPassword(boundary.getSaltAndHash());
+        User newUser = userStore.create(user, userPassword);
+
+        return new RegisterUserResponse(newUser, userPassword);
+    }
+
+    /**
+     * Registers a new {@link User}
+     * @param boundary <T> {@link IUserRegisterInputBoundary} providing input
+     * @return a new {@link RegisterUserResponse} with created {@link User}, or an error message
+     */
+    public RegisterUserResponse register(T boundary) {
+        // check email and username are not taken
+        RegisterUserResponse emailUsernameTaken = areEmailAndUsernameTaken(boundary);
+        if (emailUsernameTaken != null) {
+            return emailUsernameTaken;
+        }
+
+        // Check user input valid
+        RegisterUserResponse inputInvalid = isValidated(boundary);
+        if (inputInvalid != null) {
+            return inputInvalid;
+        }
+
+        // register a new user
+        return registerUserToDatabase(boundary);
+    }
+}
--- a/backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
+++ b/backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
@@ -1,11 +1,10 @@
 package com.javahelp.backend.domain.user.register.client;
 
 import com.javahelp.backend.data.IUserStore;
-import com.javahelp.backend.domain.user.register.RegisterUserResponse;
+import com.javahelp.backend.domain.user.register.UserRegisterResult;
 import com.javahelp.backend.domain.user.register.UserRegisterInteractor;
 import com.javahelp.model.user.ClientUserInfo;
 import com.javahelp.model.user.User;
-import com.javahelp.model.user.UserPassword;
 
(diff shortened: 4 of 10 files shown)


backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Generate a new pseudo {@link User}. Private because this is exclusively for a general
+     * {@link User} type.
+     *
+     * @return a new {@link User} entity
+     */
+    private User register(T boundary){
+        return new User("", this.bundleInfo(boundary), boundary.getUsername());
```
Jacob4649, line 32: There is no need for a method that just calls the `User` constructor. This can just be a line of code, it is only called once, and does not really add any functionality or readability IMHO.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+                    "consider logging in instead?");
+        }
+        if(IUserStore.getDefaultImplementation().readByEmail(boundary.getEmailAddress()) != null){
+            return new RegisterUserResponse(null, null, "The email" +
+                    " address is already occupied. If that is previously yours, consider logging " +
+                    "in instead?");
+        }
+        if(IUserStore.getDefaultImplementation().readByUsername(boundary.getUsername()) != null){
+            return new RegisterUserResponse(null, null, "The username" +
+                    " is already occupied. If that is previously yours, consider logging in instead?");
```
Jacob4649, line 57: These should use the instance of `IUserStore` belonging to the class, rather then getting the default implementation here.

Jacob4649, line 57: These error messages should communicate what went wrong, including suggestions like 'if that is previously yours consider logging in instead' is probably not what we want to do. IMHO its better to have a simple error message like 'Username already taken' and if the frontend wants to show a message to the end user, it can figure that out itself.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Checking for violations for registering a {@link User} account.
+     * @param boundary an instance of {@link IUserRegisterInputBoundary} to retrieve input info
+     *
+     * @return a {@link RegisterUserResponse} model that contains message of failure at registration
+     * Null if all tests passed.
+     */
+    protected RegisterUserResponse isValidRegistering(T boundary){
```
Jacob4649, line 42: This method should have a more specific name, if it checks for email and username it should be called something like `emailAndUsernameTaken`, etc...

Jacob4649: I also don't see a specific need for this to be protected rather than private

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * A method for future different types of {@link User} that might need to validate
+     * their own defined attributes (i.e. beyond the uniqueness check in username and email address)
+     * @param boundary <T> interface extends {@link IUserRegisterInputBoundary} to retrieve input
+     *                 info
+     *
+     * @return true if assumptions for registering that type of {@link User} holds; false otherwise
+     */
+    protected boolean isValidated(T boundary){
```
Jacob4649, line 72: This method could act like the `isValidRegistering` method, and return a response with an error, or else null. This way it could communicate what exactly went wrong. Additionally, this method should not call `isValidRegistering`, it should be abstract, and not define a body. This way implementing subclasses will be forced to override it and define a validation behaviour for their use cases input boundaries. Both this and `isValidRegistering` should have to return null for the use case to proceed.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+        /* I think this should be overridden in case to add some further checks for future user types.
+        isValidRegistering is "good enough" for clients and provider for now*/
+        return isValidRegistering(boundary) == null;
+    }
+
+    /**
+     * A method to bundle up {@link UserInfo} based on their given
+     * {@link com.javahelp.model.user.UserType}
+     */
+    protected U bundleInfo(T boundary){
```
Jacob4649, line 82: This method should be abstract and have a more specific name

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @return a {@link RegisterUserResponse} that contains all result info.
+     */
+    protected RegisterUserResponse registering(T boundary, User user) {
+        // Make a new user account as well as its password
+        User newUser = userStore.create(user, boundary.getPassword());
+        UserPassword password = userStore.readPassword(newUser.getStringID());
+
+        return new RegisterUserResponse(newUser, password, null);
+    }
```
Jacob4649, line 102: this method does not need to be protected, and should have a more specific name

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,126 @@
...
+        if (!result){
+            // Early return for failure. Call isValidRegistering to get a full response model
+            return this.isValidRegistering(boundary);
+        }
+
+        // After all checks passed, time to make a new user
+        User user = register(boundary);
+
+        return registering(boundary, user);
+    }
```
Jacob4649, line 125: perhaps this method should be called `register` or something. Also, the structure of calling `isValidated` then if it returns false, calling `isValidRegistering` should  be changed. `isValidRegistering` should be called. If it comes back non null, its output should be returned. The same should then be done with `isValidated`.

Jacob4649: Additionally, you don't need to qualify every use of a method or class field with `this`. It might be best practice to only do so when it remove ambiguity and is necessary for the code. Most modern IDEs display local and class fields differently so it does not really add to readability.

backend/src/main/java/com/javahelp/backend/domain/user/register/client/IClientRegisterInputBoundary.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * @return last name from the ClientRequestModel
+     */
+    String getLastName();
+
+    /**
+     * @return the type of that {@link com.javahelp.model.user.User}
+     */
+    UserType getType();
```
Jacob4649, line 31: Every `UserInfo` coming out of here will have the same `UserType` so this is unnecessary.

backend/src/main/java/com/javahelp/backend/domain/user/register/provider/IProviderRegisterInputBoundary.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * @return the certification status from the ProviderRequestModel. Return false by default
+     */
+    boolean getCertified();
+
+    /**
+     * @return the type of that {@link com.javahelp.model.user.User}
+     */
+    UserType getType();
```
Jacob4649, line 31: Also unnecessary, see above.

backend/src/test/java/com/javahelp/backend/register/ClientRegisterTest.java
```diff
@@ -0,0 +1,168 @@
...
+    @Test(timeout = 5000)
+    public void testFailure(){
+        assertNotNull("The email address already exists", c2.getError());
+    }
+
+    @Test(timeout = 5000)
+    public void testSuccess(){
+        assertNull(c1.getError());
+        assertNull(c3.getError());
+    }
```
Jacob4649, line 166: These test do not pass right now.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
```diff
@@ -10,7 +10,7 @@ public class ProviderUserInfo extends UserInfo {
     private String phoneNumber;
 
     /**
-     * Constructs a ProviderInfo object.
+     * Constructs a ProviderInfo object without given gender
```
Jacob4649, line 13: Gender should be removed from this PR entirely, it should be entirely in its own PR.

Jacob4649 (changes requested):
This PR also has a lot of commits, we would probably want to go through these before merging it.

backend/src/test/java/com/javahelp/backend/register/ClientRegisterTest.java
Lam207: You're right.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
Lam207: Sure

backend/src/main/java/com/javahelp/backend/domain/user/register/IUserRegisterInputBoundary.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * @return the username from the (Client/Provider) RequestModel
+     */
+    String getUsername();
+
+    /**
+     * @return the hashed password from the (Client/Provider) RequestModel
+     */
+    UserPassword getPassword();
```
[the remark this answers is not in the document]
Lam207: I see, so the `randomSalt()` should be used in frontend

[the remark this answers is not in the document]
Jacob4649: yes, but not directly, the frontend will construct a `UserPassword` by using the `UserPassword(String password, IPasswordHasher hasher)` constructor, and then send the base 64 encoded salt and hash.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
Lam207: You're exactly right

backend/src/main/java/com/javahelp/backend/domain/user/register/client/IClientRegisterInputBoundary.java
Lam207: Agree, though we also agreed on keeping it for future usage.

model/src/main/java/com/javahelp/model/user/ProviderUserInfo.java
Lam207: Sorry this is a misleading docstring

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
Lam207: I'm not sure but I could not make them `private`. If not `protected`, then it should be `public`, but this is only reserved after checking all input for a successful case. Therefore, I'm kind of stuck there and still keep `protected`. Let me know if there is some deeper reason that I am not aware of 👍

backend/src/main/java/com/javahelp/backend/domain/user/register/client/IClientRegisterInputBoundary.java
Jacob4649: It should not be in either, the subclass of `UserRegisterInteractor` used should only process a single type of `User` so this is not necessary anywhere.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
Jacob4649: there is a different public entry point for this logic, I am not sure why this would need to be either public or protected. Would this ever change in subclasses or need to be accessed from outside the class?

backend/src/main/java/com/javahelp/backend/domain/user/register/client/IClientRegisterInputBoundary.java
Lam207: Gotcha

backend/src/main/java/com/javahelp/backend/domain/user/register/IUserRegisterInputBoundary.java
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * @return the input username from the UserRequestModel
+     */
+    String getUsername();
+
+    /**
+     * @return the Base64-generated string salt combined with hash.
+     */
+    String getSaltandHash();
```
Jacob4649, line 21: Remember to use camelCase, `getSaltandHash` is equivalent to 'get saltand hash' in english

backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * For bundle up info of the result of this successful registration
+     * @param user the newly created {@link User}
+     * @param password the newly created {@link UserPassword}
+     */
+    public RegisterUserResponse(User user, UserPassword password){
+        this.user = user;
+        this.password = password;
+        this.error = null;
```
Jacob4649, line 20: it's not necessary to set error here, it will be null without this line. If you make it final (which you could), then this would be necessary, but as is it is not needed. That doesn't mean it needs to be fixed, I'm just letting you know for the future.

backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * An overloaded constructor to make a report for this failed registration
+     * @param error a description of failure at registration
+     */
+    public RegisterUserResponse(String error){
+        this.user = null;
+        this.password = null;
```
Jacob4649, line 29: See above comments

backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public String getErrorMessage() {
+        return error;
+    }
+
+    /**
+     * @return an instance of a newly created {@link User}
+     * Null if failed; success otherwise
+     */
+    public User getUser(){
```
Jacob4649, line 45: There should be a space between the brackets and the curly bracket

backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
```diff
@@ -0,0 +1,70 @@
...
+        return user;
+    }
+
+    /**
+     * @return an instance of a newly made {@link UserPassword}
+     * Null if failed, success otherwise
+     */
+    public UserPassword getPassword(){
```
Jacob4649, line 53: You might want to run the formatter on your code. See the above comment about spacing.

backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
```diff
@@ -0,0 +1,70 @@
...
+    private User user;
+    private UserPassword password;
+    private String error;
+
+    /**
+     * For bundle up info of the result of this successful registration
+     * @param user the newly created {@link User}
+     * @param password the newly created {@link UserPassword}
+     */
+    public RegisterUserResponse(User user, UserPassword password){
```
Jacob4649, line 17: These constructors should probably not be public. You want to be able to interact with this class publicly, but the only time a new instance should be created, is by the interactor, which is in the same package. These should keep default visibility.

backend/src/main/java/com/javahelp/backend/domain/user/register/RegisterUserResponse.java
```diff
@@ -0,0 +1,70 @@
...
+    public void setUser(User newUser){
+        this.user = newUser;
+    }
+
+    /**
+     * @param newPassword an updated password
+     */
+    public void setPassword(UserPassword newPassword){
+        this.password = newPassword;
+    }
```
Jacob4649, line 69: This class should not have a setter for either of these fields. This is for the same reason as the constructor visibility, there is no scenario where either of these fields would be set outside of the constructor.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,80 @@
+package com.javahelp.backend.domain.user.register;
+
+import com.javahelp.backend.data.IUserStore;
+import com.javahelp.model.user.User;
+import com.javahelp.model.user.UserInfo;
+import com.javahelp.model.user.UserPassword;
+
+/**
+ * An interactor to make new {@link User} and {@link UserPassword} for all types of users
```
Jacob4649, line 9: The generic type parameters should be documented here

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,80 @@
...
+        if ((userStore.readByEmail(boundary.getEmailAddress()) != null)
+                && (userStore.readByUsername(boundary.getUsername()) != null)){
+            // If both username and email address exist, then return both errors
+            return new RegisterUserResponse("Both the given email address and username " +
+                    "are already occupied");
+        }
+        if (IUserStore.getDefaultImplementation().readByEmail(boundary.getEmailAddress()) != null){
+            return new RegisterUserResponse("The email address is already occupied");
+        }
+        else if (IUserStore.getDefaultImplementation().readByUsername(boundary.getUsername()) != null){
```
Jacob4649, line 42: You shouldn't be suing `IUserStore.getDefaultImplementation` here. You should be using the class' `userStore` field. Also, instead of running these queries 4 times, you can just run them twice and store the results from the first two. There is no reason to fetch the same information twice for each of these, it will slow down the process by requiring twice as many database calls as necessary.

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
```diff
@@ -0,0 +1,80 @@
...
+     * An abstract method to create a new user entity for this user based on their info,
+     * which will be defined in other subtypes of users.
+     * @param boundary <T> interface extends {@link IUserRegisterInputBoundary} to get info from the
+     *                 register View Model.
+     *
+     * @return a new {@link RegisterUserResponse} entity that comprises of {@link User}
+     * {@link UserPassword} if success
+     * OR of only a {@link String} error message if failed.
+     */
+    protected abstract RegisterUserResponse register(T boundary);
```
Jacob4649, line 79: This method should not be abstract, it has a very well defined behaviour, that will be the same in every implementation. To pass the created `User` object, with the `UserInfo` created in `bundleUserInfo` along with the created `UserPassword` object to `userStore.create`. This behaviour is the same for every single implementation so making this abstract doesn't make much sense.

backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
```diff
@@ -0,0 +1,81 @@
...
+import com.javahelp.model.user.User;
+import com.javahelp.model.user.UserPassword;
+
+/**
+ * An interactor that creates new {@link User} and {@link UserPassword} for this client
+ */
+public class ClientRegisterInteractor extends
+        UserRegisterInteractor<IClientRegisterInputBoundary, ClientUserInfo> {
+
+    IUserStore userStore;
```
Jacob4649, line 16: There is no need to duplicate this field, the superclass already has it. Putting it here again hides the superclass field.

backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
```diff
@@ -0,0 +1,81 @@
...
+            return result;
+        }
+
+        // After all checks passed, time to make a new client
+        User user = registerClient(boundary);
+
+        UserPassword userPassword = new UserPassword(boundary.getSaltandHash());
+        User newUser = userStore.create(user, userPassword);
+
+        return new RegisterUserResponse(newUser, userPassword);
```
Jacob4649, line 51: This implementation has nothing to do with clients specifically. It works for very kind of `User` and should just be in the superclass.

backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    @Override
+    protected ClientUserInfo bundleUserInfo(IClientRegisterInputBoundary boundary){
+        return new ClientUserInfo(boundary.getEmailAddress(), boundary.getAddress(),
+                boundary.getPhoneNumber(), boundary.getFirstName(), boundary.getLastName());
+    }
+
+    @Override
+    protected RegisterUserResponse isValidated(IClientRegisterInputBoundary boundary) {
+        return super.areEmailAndUsernameTaken(boundary);
```
Jacob4649, line 68: Instead of calling this in this implementation, this method should be called somewhere in the superclass, before `isValidated` is called

backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * A helper method to register a client {@link User}
+     * @param boundary the given {@link IClientRegisterInputBoundary} to get the input info from
+     *                 frontend.
+     * @return a new {@link User} for the Client
+     */
+    protected User registerClient(IClientRegisterInputBoundary boundary){
+        ClientUserInfo Cinfo = bundleUserInfo(boundary);
+        return new User("", Cinfo, boundary.getUsername());
+    }
```
Jacob4649, line 80: This method seems unnecessary to me, it just calls the constructor for `User`

backend/src/main/java/com/javahelp/backend/domain/user/register/provider/ProviderRegisterInteractor.java
```diff
@@ -0,0 +1,83 @@
...
+import com.javahelp.model.user.User;
+import com.javahelp.model.user.UserPassword;
+
+/**
+ * An interactor that registers new {@link User} and {@link UserPassword} for this provider
+ */
+public class ProviderRegisterInteractor extends
+        UserRegisterInteractor<IProviderRegisterInputBoundary, ProviderUserInfo> {
+
+    IUserStore userStore;
```
Jacob4649, line 16: See comments on `ClientRegisterInteractor`, this field should not be duplicated

Jacob4649 (changes requested):

backend/src/main/java/com/javahelp/backend/domain/user/register/UserRegisterInteractor.java
Lam207: Oh my bad, thanks for the reminder!

Jacob4649 (approved):
Looks good!

BullDF (approved):
Looking good!

backend/src/main/java/com/javahelp/backend/domain/user/register/client/ClientRegisterInteractor.java
```diff
@@ -0,0 +1,34 @@
...
+    protected ClientUserInfo bundleUserInfo(IClientRegisterInputBoundary boundary) {
+        return new ClientUserInfo(boundary.getEmailAddress(), boundary.getAddress(),
+                boundary.getPhoneNumber(), boundary.getFirstName(), boundary.getLastName());
+    }
+
+    @Override
+    protected UserRegisterResult isValidated(IClientRegisterInputBoundary boundary) {
+        return null;
+    }
+}
```
Enkhzul001, line 34: Why does this return nulll?

Enkhzul001 (approved):
Overall, looks great! Good Work

PR #103, opened 2022-11-21
Login frontend

### LINKED ISSUE #91: [Frontend | ViewModel| Login ViewModel
Login viewmodel

Jacob4649:
Closes #91.

### DIFF
--- a/app/src/main/java/com/javahelp/frontend/domain/user/login/LoginInteractor.java
+++ b/app/src/main/java/com/javahelp/frontend/domain/user/login/LoginInteractor.java
@@ -9,7 +9,7 @@ import org.apache.hc.core5.concurrent.FutureCallback;
 /**
  * Use case interactor for logging in a {@link User}
  */
-class LoginInteractor implements ILoginInput {
+public class LoginInteractor implements ILoginInput {
 
     ILoginOutput output;
     ISaltDataAccess saltAccess;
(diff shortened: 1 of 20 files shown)


Lenzs (approved):
looks really good!

app/src/main/java/com/javahelp/frontend/domain/user/login/LoginInteractor.java
```diff
@@ -0,0 +1,93 @@
...
+        this.output = output;
+        this.saltAccess = saltAccess;
+        this.loginAccess = loginAccess;
+        this.hasher = hasher;
+    }
+
+    @Override
+    public void login(String id, String username, String email, String password, boolean stayLoggedIn) {
+
+        if (username == null && email == null && id == null) {
```
Enkhzul001, line 38: Is Id the token id?

Enkhzul001 (approved):
Everything looks good.

app/src/main/java/com/javahelp/frontend/domain/user/login/LoginInteractor.java
Jacob4649: ID here is the ID of the user to log into.  One of `id`, `username`, and `email` must be non-null. The other two can be null. If all three are null, an error is thrown. This allows users to login using a variety of different pieces of identifying information. If we wanted to change it so that in the future users could log in by their email, we could. Similarly, if we wanted users to be able to save their user IDs and log in faster by just entering their password, we could!

PR #107, opened 2022-11-22
Survey dynamo gateway

stevolopolis:


backend/src/main/java/com/javahelp/backend/data/ISurveyResponseStore.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param constraint {@link Map <>} which specifies the required attributes
+     *                   from the User.
+     * @return {@link Set <User>} with the specified constraints.
+     */
+    Set<String> readByConstraint(Map<String, Set<String>> constraint);
+
+    /**
+     *
+     * @return all {@link Set<User>} existing in database.
```
Jacob4649, line 64: Javadocs return type doesn't match function return type

backend/src/main/java/com/javahelp/backend/data/ISurveyStore.java
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @param id {@link String} id of the {@link Survey} to delete
+     */
+    void delete(String id);
+
+    /**
+     * Removes all {@link Survey}s in database.
+     * ONLY use during preliminary testing!
+     */
+    void cleanTable();
```
Jacob4649, line 39: Is the  clean table method  meant to still be here?

backend/src/test/java/com/javahelp/backend/data/DynamoDBSurveyResponseStoreTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    Regions regions = Regions.US_EAST_1;
+
+    DynamoDBSurveyResponseStore db = new DynamoDBSurveyResponseStore(tableName, regions);
+
+    @Test(timeout = 5000)
+    public void testCreateRead() {
+        this.db.cleanTable();
+
+        User user = setupUser();
```
Jacob4649, line 33: you can add 
```java

private boolean databaseAccessible() {
    try {
        this.db.read("test");
        return true;
    } catch (Exception e) {
        return false;
    }
}

// ... now inside test method ...

assumeTrue(databaseAccessible());
```

To tests to prevent them from running unless the condition to assume returns true. This would prevent these tests from running unless they have access to the database.

backend/src/test/java/com/javahelp/backend/data/DynamoDBSurveyResponseStoreTest.java
```diff
@@ -0,0 +1,131 @@
...
+        SurveyResponse sr = setupSurveyResponse();
+
+        sr = this.db.create(user.getStringID(), sr);
+
+        SurveyResponse read = this.db.read(sr.getID());
+
+        testSurveyEqual(sr.getSurvey(), read.getSurvey());
+        testResponsesEqual(sr.getSurvey(), sr, read);
+
+        this.db.delete(sr.getID());
```
Jacob4649, line 43: Similarly, it might be wise to enclose this in the `finally` block of a `try`-`finally` for guaranteed resource cleanup

backend/src/main/java/com/javahelp/backend/data/DynamoDBSurveyResponseStore.java
```diff
@@ -0,0 +1,266 @@
...
+     * @param userID {@link String} id of the user to be queried.
+     * @return {@link List <SurveyResponse>} of the specified user.
+     */
+    @Override
+    public List<SurveyResponse> readByUser(String userID) {
+        String keyConditionExpression = "user_id = :id";
+        Map<String, AttributeValue> expressionAttributeValues = new HashMap<>();
+        expressionAttributeValues.put("id", new AttributeValue().withS(userID));
+
+        ScanRequest scanRequest = new ScanRequest()
```
Lam207, line 128: I'm not really sure where `ScanRequest` comes from, but this is guaranteed not to be `null`, right?

Lam207 (commented):
Amazing work so far!

backend/src/main/java/com/javahelp/backend/data/DynamoDBSurveyResponseStore.java
```diff
@@ -0,0 +1,266 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+
+
+public class DynamoDBSurveyResponseStore extends DynamoDBStore implements ISurveyResponseStore{
+    private String tableName;
+    private static ISurveyStore surveyStore = ISurveyStore.getDefaultImplementation();
```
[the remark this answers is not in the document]
stevolopolis: Are you referring to line 28, where I make use of SurveyStore?

backend/src/main/java/com/javahelp/backend/data/DynamoDBSurveyResponseStore.java
```diff
@@ -0,0 +1,266 @@
...
+    /**
+     * Creates a {@link SurveyResponse}
+     *
+     * @param userID {@link String} id of author (user) of this surveyResponse
+     * @param surveyResponse {@link SurveyResponse} to create
+     * @return {@link SurveyResponse} that was created
+     */
+    @Override
+    public SurveyResponse create(String userID, SurveyResponse surveyResponse) {
+        surveyStore.cleanTable();
```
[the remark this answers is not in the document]
stevolopolis: Removed. Was included for preliminary testing, but should not be included in final code.

backend/src/main/java/com/javahelp/backend/data/ISurveyResponseStore.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    List<SurveyResponse> readByUser(String userID);
+
+    /**
+     *
+     * @param constraint {@link Map <>} which specifies the required attributes
+     *                   from the User.
+     * @return {@link Set <User>} with the specified constraints.
+     */
+    Set<String> readByConstraint(Map<String, Set<String>> constraint);
```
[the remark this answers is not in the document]
stevolopolis: I agree, but I also think the current specified type for the constraint is quite universal as it simply specifies the set of possible solutions to a given query constraint. How that query constraint is satisfied is not specified in the interface. Do correct me if I'm wrong

backend/src/main/java/com/javahelp/backend/data/ISurveyStore.java
stevolopolis: I think so, at least for now. Otherwise there is no other way to quickly reset the table of Surveys.

backend/src/test/java/com/javahelp/backend/data/DynamoDBSurveyResponseStoreTest.java
stevolopolis: Good point

backend/src/main/java/com/javahelp/backend/data/DynamoDBSurveyResponseStore.java
stevolopolis: The null case is handled in the for loop below this line.

backend/src/main/java/com/javahelp/backend/data/ISurveyResponseStore.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.amazonaws.regions.Regions;
+import com.javahelp.model.survey.SurveyResponse;
+import com.javahelp.model.user.User;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public interface ISurveyResponseStore {
```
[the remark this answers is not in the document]
stevolopolis: Agreed. Will be adding an update method.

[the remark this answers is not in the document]
stevolopolis: Actually, since SurveyResponse class doesn't have setters for either the survey objects or the responses objects, I don't see the purpose of adding an update method. Rather, the easier way to implement the idea is to always delete prior surveys created by the same user before adding a new survey entry.

What do you think?

backend/src/test/java/com/javahelp/backend/data/DynamoDBSurveyResponseStoreTest.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.Map;
+
+public class DynamoDBSurveyResponseStoreTest {
+    String tableName = "javahelpBackendFilledSurveys";
+
+    Regions regions = Regions.US_EAST_1;
+
+    DynamoDBSurveyResponseStore db = new DynamoDBSurveyResponseStore(tableName, regions);
+
+    @Test(timeout = 5000)
```
Enkhzul001, line 30: Is the timeout suppose to be this long in this case?

Enkhzul001 (approved):
Overall, looks good

backend/src/test/java/com/javahelp/backend/data/DynamoDBSurveyResponseStoreTest.java
Lam207: For Dynamo, I think 5s is good yeah.

model/src/main/java/com/javahelp/model/survey/SurveyResponse.java
```diff
@@ -42,10 +42,7 @@ public Survey getSurvey() {
      * @return {@link SurveyQuestionResponse} for the provided {@link SurveyQuestion}
      */
     public SurveyQuestionResponse getResponse(SurveyQuestion question) {
-        if (!responses.containsKey(question)) {
-            throw new IllegalArgumentException("The SurveyQuestion provided has no response");
-        }
-        return responses.get(question);
+        return responses.getOrDefault(question, new SurveyQuestionResponse(question, -1));
```
BullDF, line 45: I like that you use this method in `Map` to simplify the if-else statement.

BullDF: However, I think it's better to specify in the Java doc for this method what -1 means, as I was confused whether -1 means no question found or there is such a question but no response.

backend/src/main/java/com/javahelp/backend/data/ISurveyStore.java
```diff
@@ -0,0 +1,40 @@
+package com.javahelp.backend.data;
+
+import com.amazonaws.regions.Regions;
+import com.javahelp.model.survey.Survey;
+
+public interface ISurveyStore {
```
BullDF, line 6: Do we need this to extend the `ICRUDGateway` interface? (I think Jacob and you had this conversation before but I don't remember the specifics)

BullDF (commented):
Good stuff Steven. Just not sure about the `update()` method in the database. What do you think?

backend/src/main/java/com/javahelp/backend/data/ISurveyResponseStore.java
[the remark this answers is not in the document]
Lam207: I personally think your approach of making a new survey each time sounds better, because each survey should have a different `id` with it, even if it was done by the same `User`. 

I feel like the `SurveyResponse` was implemented that way to keep a result version distinct from other survey answers.

model/src/main/java/com/javahelp/model/survey/SurveyResponse.java
Lam207: I think the second `int` parameter is used as a selected index in the whole survey (which is like a collection) to extract that specific response based on the given index. Not completely sure though. It was previously set as `0`, and I think `-1` mean that the index is invalid, as it is out of the scope of the collection `SurveyQuestion`

backend/src/main/java/com/javahelp/backend/data/DynamoDBSurveyResponseStore.java
```diff
@@ -0,0 +1,367 @@
...
+                n_answers = n_answers + 1;
+            }
+            String attrKeyString = String.format("responses.%s", formattedQuestion);
+            String attrKeySetString = String.format("(%s)", String.join(", ", expressions));
+            conditionBuilder.append("(");
+            conditionBuilder.append(attrKeyString);
+            conditionBuilder.append(" IN ");
+            conditionBuilder.append(attrKeySetString);
+            conditionBuilder.append(")");
+            if (n_questions < constraint.size()) {
```
Lam207, line 236: Good thought on when to add `"AND"`!

Lam207 (commented):
Good work! Though, there are still some follow-up suggestions

model/src/main/java/com/javahelp/model/survey/SurveyResponse.java
stevolopolis: `-1` is just a number I chose to represent situations when this question is not answered by the User. Then, for example, whenever the providerRanker detects `-1` as an answer, it will know not to compare the values. However, I do agree that adding docs is good.

backend/src/main/java/com/javahelp/backend/data/ISurveyStore.java
stevolopolis: I extended ISurveyResponse but not SurveyStore. Not the most urgent thing to change but will change it

backend/src/main/java/com/javahelp/backend/data/ISurveyResponseStore.java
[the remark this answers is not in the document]
Jacob4649: I agree with this, there does not seem to be much use in an update method

backend/src/main/java/com/javahelp/backend/data/ISurveyResponseStore.java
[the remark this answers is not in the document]
Jacob4649: This is true; however the way the input is currently setup, arbitrary strings will have to be parsed into corresponding field/column names depending on the underlying implementation. Additionally, the constraints will need to be parsed from strings into queries for the database. This essentially amounts to specifying our own query language, and then implementing that query language over some underlying database. This adds a lot of complexity, and since the method takes in strings, makes future refactors much more difficult. While this is definitely a functioning way to do this, it could lead to future issues. Either way, although there are better ways to do it, I think it's good enough for our purposes. If we really want to we can mark it as an area for future improvement in our submission,

Jacob4649 (approved):
Overall, looks good

Lam207 (approved):
Good work!

PR #110, opened 2022-11-26
Add homepage UI

Enkhzul001:
I added homepage, account page, and bottom navigation view. I left search page part for now. I arranged registration form and linked registration button in LoginActivity to their respective activities: Client or Provider registration.


.idea/assetWizardSettings.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Jacob4649, line 1: This file probably shouldn't have been committed.

app/build.gradle
```diff
@@ -24,6 +24,10 @@ android {
         }
     }
 
+    buildFeatures{
+        viewBinding true;
+    }
```
Jacob4649, line 29: I think this should already have been in the gradle, its probably best to avoid messing with the gradle files.

Jacob4649: Just double checked, and it is indeed already in the gradle on main.

app/src/main/AndroidManifest.xml
```diff
@@ -25,17 +25,15 @@
         <activity
             android:name=".frontend.activity.ClientRegistrationActivity"
             android:exported="false"
-            android:parentActivityName=".frontend.activity.LoginActivity"
-            android:theme="@style/Theme.AppCompat.NoActionBar">
```
Jacob4649, line 29: We might want to leave the no action bar in

app/src/main/java/com/javahelp/frontend/activity/LoginActivity.java
```diff
@@ -58,6 +59,19 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            public void onClick(View v) {
+                startActivity(new Intent(LoginActivity.this, ClientRegistrationActivity.class));
+            }
+        });
+        binding.botPReg.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                startActivity(new Intent(LoginActivity.this, ProviderRegistrationActivity.class));
+            }
+        });
```
Jacob4649, line 74: These should be done with lambda expressions or method references. See line 61 for an example of a method reference, and I think there should be some lambda expressions further up in the file. It's just a lot more readable.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,6 +26,10 @@ protected void onCreate(Bundle savedInstanceState) {
         binding.setPdata(providerRegistrationViewmodel);
         binding.setLifecycleOwner(this);
 
+        //setting up spinner to create dropdown item for gender
+        ArrayAdapter<CharSequence> adapter=ArrayAdapter.createFromResource(this, R.array.genders, android.R.layout.simple_spinner_item);
+        adapter.setDropDownViewResource(android.R.layout.simple_spinner_item);
+        binding.gender.setAdapter(adapter);
```
Jacob4649, line 32: Since it's used in several places, perhaps we could make a `GenderSelector` `View`.

app/src/main/java/com/javahelp/frontend/fragments/AccountFragment.java
```diff
@@ -0,0 +1,66 @@
...
+public class AccountFragment extends Fragment {
+
+    // TODO: Rename parameter arguments, choose names that match
+    // the fragment initialization parameters, e.g. ARG_ITEM_NUMBER
+    private static final String ARG_PARAM1 = "param1";
+    private static final String ARG_PARAM2 = "param2";
+
+    // TODO: Rename and change types of parameters
+    private String mParam1;
+    private String mParam2;
```
Jacob4649, line 27: Comments should be added to make it clear that this is example code & boilerplate, and that this will be removed and filled in with content later.

Jacob4649 (changes requested):
A few small things, but overall, I think it looks really good!

app/src/main/AndroidManifest.xml
Enkhzul001: Can you explain why? We are on Client registration activity so wouldn't we want to go back to login page if we don't want register?

app/src/main/java/com/javahelp/frontend/fragments/AccountFragment.java
Enkhzul001: ok

app/src/main/java/com/javahelp/frontend/activity/LoginActivity.java
Enkhzul001: That is actually good idea

app/build.gradle
Enkhzul001: Ah I see. I fix this part on my new commit.

Lam207 (commented):
Looks good to me with full coverage on input data

Jacob4649 (approved):
Looks good!

app/src/main/java/com/javahelp/frontend/fragments/AccountFragment.java
```diff
@@ -0,0 +1,62 @@
...
+        args.putString(ARG_PARAM1, param1);
+        args.putString(ARG_PARAM2, param2);
+        fragment.setArguments(args);
+        return fragment;
+    }
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        if (getArguments() != null) {
```
Lam207, line 50: I think it is OK not to worry about when `getArguments` from `Fragment` superclass to return `null`, because it is difficult to handle that.

Lam207 (approved):
Looks good to me!

Jacob4649 (approved):
Looks great!

Lenzs (approved):
everything looks good!

PR #111, opened 2022-11-28
[Backend] Account Deletion Handler

BullDF:
The handler for an account deletion.

### DIFF
--- a/backend/src/main/java/com/javahelp/backend/endpoint/HTTPTokenHandler.java
+++ b/backend/src/main/java/com/javahelp/backend/endpoint/HTTPTokenHandler.java
@@ -59,7 +59,7 @@ public abstract class HTTPTokenHandler extends HTTPHandler implements ITokenAuth
         TokenAuthResult result = interactor.authenticate(this);
 
         if (result.isSuccess()) {
-            return authenticatedGetResponse(result.getUser(), result.getToken(), body, method, headers, parameters);
+            return authenticatedGetResponse(result.getUser(), result.getToken(), body, method, headers, parameters, pathParameters);
         } else {
             return APIGatewayResponse.error(UNAUTHENTICATED, result.getErrorMessage()); // invalid credentials
         }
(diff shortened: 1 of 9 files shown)


Jacob4649 (approved):
Looks great!

yuanshu777 (approved):
It looks great.

backend/src/main/java/com/javahelp/backend/endpoint/HTTPTokenHandler.java
```diff
@@ -99,8 +100,8 @@ private boolean extractTokenAndUserID(String[] authHeader) {
         for (String s : authHeader) {
             if (s.startsWith("id=")) {
                 userId = s.substring(3);
-            } else if (s.startsWith("username=")) {
-                tokenString = s.substring(9);
+            } else if (s.startsWith("token=")) {
+                tokenString = s.substring(6);
```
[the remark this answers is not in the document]
BullDF: When we passed the information into the handler it comes in the form of a HTTP header (as you can see in the tests), where the ID for the token is specified after `token=` in the header (there are 6 letters/= in `token=` so the actual ID starts at index 6). It was 9 previously because Jacob accidentally typed `username=` (where there are 9 letters/= in it) so I was fixing that typo.

backend/src/test/java/com/javahelp/backend/endpoint/delete/DeleteHandlerTest.java
```diff
@@ -0,0 +1,113 @@
...
+        try {
+            userStore.read("test");
+            tokenStore.read("test");
+            return true;
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    @Test
```
Lam207, line 53: Um I don't know whether this is necessary but should there be a limit in timeout like 10000 (10s) or 50s at max?

backend/src/test/java/com/javahelp/backend/endpoint/delete/DeleteHandlerTest.java
```diff
@@ -0,0 +1,113 @@
...
+                    .when().delete(USER_ENDPOINT + user.getStringID()).then().statusCode(200)
+                    .body("success", equalTo(true));
+
+        } finally {
+            tokenStore.delete(token.getToken());
+            userStore.delete(user.getStringID());
+        }
+    }
+
+    @Test
```
Lam207, line 81: Same with the timeout constraint comment above

Lam207 (approved):
Well done!

PR #113, opened 2022-11-30
User query constraint

stevolopolis:


Lam207 (approved):
Looks good to me, some good changes, such as for `formattedQuestion`

Enkhzul001 (approved):
Looks good

backend/src/main/java/com/javahelp/backend/query/Constraint.java
```diff
@@ -0,0 +1,24 @@
...
+    public void setConstraint(String query) {
+        this.queryList.add(query);
+    }
+
+    public HashMap<String, Set<String>> getConstraint(){
+        HashMap<String, Set<String>> constraint = new HashMap<>();
+        constraint.put(this.question, this.queryList);
+        return constraint;
+    };
+}
```
Jacob4649, line 24: This should be documented before we hand this in, but not necessarily in this PR just in the interest of keeping things moving quickly

backend/src/main/java/com/javahelp/backend/query/IUserQueryConstraint.java
```diff
@@ -0,0 +1,25 @@
...
+    Map<String, User> getProvidersWithConstraints(List<Constraint> constraints);
+
+    /**
+     *
+     * @param constraints{ @link List} of {@link Constraint}s to be applied for query.
+     * @return {@link Map} of {@link String} ID of users and their corresponding
+     * {@link SurveyResponse}s.
+     */
+    Map<String, SurveyResponse> getResponsesWithConstraints(List<Constraint> constraints);
+}
```
Jacob4649, line 25: I don't see a ton of use in this interface, since it's implementation is in terms of other interfaces and abstract classes. Adding another layer of abstraction might not be needed here.

Jacob4649 (approved):
A few small things, but I think maybe they could be fixed in a future PR so we can keep moving forward with the survey stuff.

backend/src/main/java/com/javahelp/backend/query/Constraint.java
stevolopolis: Agreed

PR #121, opened 2022-12-03
Provider_User_Register Testing

yuanshu777:


app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+     * Called when the Register button is clicked
+     *
+     * @param v {@link View} that was clicked
+     */
+    private void registerClick(View v) {
+        providerRegistrationViewmodel.setUsername(binding.username.getText().toString());
+        providerRegistrationViewmodel.setPassword1(binding.password.getText().toString());
+        providerRegistrationViewmodel.setPassword2(binding.repassword.getText().toString());
+        providerRegistrationViewmodel.setProviderUserInfo(binding.email.getText().toString(),
+                binding.firstname.getText().toString(),
```
Lam207, line 69: This should be `binding.practicename.getText().toString()` and there should be (optional) one for `isCertified` for `Providers` (`false` by default)

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                binding.lastname.getText().toString(),
+                binding.home.getText().toString(),
+                binding.phonenumber.getText().toString());
+        registerAttempt();
+    }
+
+    /**
```
Lam207, line 76: I like the way this method is implemented. Though, for Javadoc, in my opinion, it should be more specific like `Initiates a register request, while setting error descriptions to {@link providerRegistrationViewModel} if there is any violation in input`.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+     */
+    private void registerAttempt() {
+        if (binding.username.getText().toString().isEmpty() || binding.password.getText().toString().isEmpty()
+        || binding.repassword.getText().toString().isEmpty()) {
+            providerRegistrationViewmodel.setRegisterError("Please enter username and password");
+            return; // early return on invalid username or password
+        }else if(!binding.password.getText().toString().equals(binding.repassword.getText().toString())){
+            providerRegistrationViewmodel.setRegisterError("Passwords do not match");
+            return;
+            }else if(binding.email.getText().toString().isEmpty() || binding.firstname.getText().toString().isEmpty()
```
Lam207, line 87: This should be changed for `Providers` into `practiceName` and optional `isCertified`

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+     * @param token {@link Token} to store
+     */
+    private void storeCredentials(User user, Token token) {
+        SharedPreferencesAuthInformationProvider credentialStore =
+                new SharedPreferencesAuthInformationProvider(this);
+        try {
+            credentialStore.setTokenString(token.getToken());
+            credentialStore.setUserID(user.getStringID());
+        } catch (NoSuchPaddingException | IllegalBlockSizeException | NoSuchAlgorithmException | BadPaddingException | InvalidKeyException ignored) {
+            Toast.makeText(this, "Error Storing Credentials", Toast.LENGTH_LONG);
```
Lam207, line 140: Great work!

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+        switch (requestCode) {
+            case REQUEST_INTERNET_REGISTER:
+                if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                    providerRegistrationViewmodel.attemptRegister();
+                } else {
+                    Toast.makeText(this, "Internet access denied, cannot log in", Toast.LENGTH_LONG);
+                }
+                return;
```
Lam207, line 154: I personally think this line 154 is not necessary.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationViewmodel.java
```diff
@@ -1,17 +1,199 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public class ProviderRegistrationViewmodel extends AndroidViewModel implements IRegisterOutput {
+
+    private MutableLiveData<String> username = new MutableLiveData<>("");
+    private MutableLiveData<String> password1 = new MutableLiveData<>("");
+    private MutableLiveData<String> password2 = new MutableLiveData<>("");
+    private MutableLiveData<String> email = new MutableLiveData<>("");
+    private MutableLiveData<String> firstname = new MutableLiveData<>("");
```
Lam207, line 35: Yeah, these two `firstname` and `lastname` should become `practicename` instead.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationViewmodel.java
```diff
@@ -1,17 +1,199 @@
...
+     */
+    public void setPassword2(String password2) { this.password2.setValue(password2);}
+
+    /**
+     * Sets the email for this {@link ProviderRegistrationViewmodel}
+     * @param email {@link String} email to use
+     */
+    public void setEmail(String email) { this.email.setValue(email);}
+
+    /**
```
Lam207, line 83: One setter for `setPracticeName` :)

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationViewmodel.java
```diff
@@ -1,17 +1,199 @@
...
+    public void setCertified(boolean c) {certified.setValue(c);}
+
+    /**
+     * Sets the register error
+     * @param s {@link String} register error to set
+     */
+    public void setRegisterError(String s) {
+        registerResult.setValue(Optional.of(new RegisterResult(s)));}
+
+    /**
```
Lam207, line 120: In the `[Backend] (Provider) Register User Use Case` will bundle these together, you don't need to worry about doing it here (and also for `Clean Architecture` I think)

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationViewmodel.java
```diff
@@ -1,17 +1,199 @@
...
+     *
+     * @return the {@link RegisterResult} of the latest register attempt, if applicable
+     */
+    public MutableLiveData<Optional<RegisterResult>> getRegisterResult() {
+        return registerResult;
+    }
+
+
+    public boolean passwordMatch(String password1, String password2){
+        if (password1.equals(password2)){
```
Lam207, line 142: This could be further simplified into `return password1.equals(password2);`

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+     * Updates the UI based on a {@link RegisterResult}. This is called with non-null register results, meaning
+     * a registration has been attempted, and has either succeeded or failed. This is unlike updateOnRegisterResult
+     * which is called with an {@link Optional} {@link RegisterResult}, and calls this method if that
+     * {@link Optional} is present.
+     *
+     * @param result {@link RegisterResult} of login attempt
+     */
+    private void updateOnPresentRegisterResult(RegisterResult result) {
+        if (result.isSuccess()) {
+            binding.registerErrorText.setText("Register successful");
```
Lam207, line 118: Based on your Javadoc description, I personally think `binding.registerErrorText` could be renamed to `bingding.registerResultText` to avoid misunderstanding.

app/src/main/java/com/javahelp/frontend/domain/user/register/IRegisterInput.java
```diff
@@ -0,0 +1,22 @@
+package com.javahelp.frontend.domain.user.register;
```
Lam207, line 1: This is already done in `[Backend] Register User Use Case`, or is there anything else specific for this on `Frontend`?

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                binding.lastname.getText().toString(),
+                binding.home.getText().toString(),
+                binding.phonenumber.getText().toString());
+        registerAttempt();
+    }
+
+    /**
+     * Initiates a register request
+     */
+    private void registerAttempt() {
```
Lam207, line 79: Excellent!

app/src/main/java/com/javahelp/frontend/domain/user/register/RegisterResult.java
```diff
@@ -0,0 +1,66 @@
+package com.javahelp.frontend.domain.user.register;
```
Lam207, line 1: I feel like `com.javahelp.backend.domain.user.register.UserRegisterResult` could be reused here, and further extended like set `Token`, etc.

Lam207 (commented):
Awesome work! Let me know if some of my suggestions have some logical problems.

Lam207 (changes requested):
For the above messages

app/src/main/java/com/javahelp/frontend/domain/user/register/IRegisterInput.java
BullDF: I think this is included again so that the frontend and backend can do the work at the same time. Don't know what the specific reasons for this but I did this for my delete account as well.

app/src/main/java/com/javahelp/frontend/domain/user/register/RegisterResult.java
BullDF: Hmmm I don't think so Lam because the frontend cannot depend on the backend so they cannot share the same class.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +50,109 @@ protected void onCreate(Bundle savedInstanceState) {
...
+     * which is called with an {@link Optional} {@link RegisterResult}, and calls this method if that
+     * {@link Optional} is present.
+     *
+     * @param result {@link RegisterResult} of login attempt
+     */
+    private void updateOnPresentRegisterResult(RegisterResult result) {
+        if (result.isSuccess()) {
+            binding.registerErrorText.setText("Register successful");
+            storeCredentials(result.getUser(), result.getToken());
+            startActivity(new Intent(ProviderRegistrationActivity.this, FrontPageActivity.class));
```
BullDF, line 120: I am not sure what is done on the backend but this means after they register, they are automatically logged in. I don't know whether there will be any token issue associated with this or not. I am thinking that after they register, they can be directed to the login page again and they will have to enter their username and password again. In this case we don't need to worry about token stuff in the register use case.

BullDF (commented):
Overall great Owen. Just having one question on the activity. Please check it out!

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
yuanshu777: the practice name is set in the ViewModel by  "this.ProviderUserInfo.setPracticeName(firstname + lastname);"
Same as the certified which is set as false by "this.providerUserInfo.setCertified(false);"
you can have a look at ProviderRegistrationViewmodel

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
yuanshu777: yes, but we do not have practicename for our registration UI, I just assume practicename is last name + first name.
Also, we do not have a certified option on the provider registration UI, and it is hard to actually implement the certified functionality, I think we just set it as false

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationViewmodel.java
yuanshu777: yes!

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
yuanshu777: Yes, but what I have discussed with Jacob is that we could automatically log in after the registration, I am not too sure about it because I have to also wait for Jacob to implement the gateway

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
Lam207: I guess yeah, we could get both `lastName` and `firstName` from the UI, then combine them both (`First` + `Last` name) together to make a `practiceName` in somewhere like `ProviderRegistrationViewmodel.java`? before sending it to the backend.

app/src/main/java/com/javahelp/frontend/domain/user/register/IRegisterInput.java
Lam207: Um Johnny, interesting thought.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
```diff
@@ -25,14 +51,123 @@ protected void onCreate(Bundle savedInstanceState) {
...
+     * {@link Optional} is present.
+     *
+     * @param result {@link RegisterResult} of login attempt
+     */
+    private void updateOnPresentRegisterResult(RegisterResult result) {
+        if (result.isSuccess()) {
+            binding.registerErrorText.setText("Register successful");
+            storeCredentials(result.getUser(), result.getToken());
+            Toast.makeText(this, "Register successful", Toast.LENGTH_SHORT);
+            startActivity(new Intent(ProviderRegistrationActivity.this, FrontPageActivity.class));
```
Enkhzul001, line 135: Since this is registering in as Provider, I think we should go to frontpage-2 activity and not frontpage because this one is client's.

app/src/main/res/layout/activity_preg.xml
```diff
@@ -167,6 +167,16 @@
...
+                <TextView
+                    android:id="@+id/registerErrorText"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:text="Register Error Message"
+                    android:textAlignment="center"
+                    android:textColor="@color/black"
+                    android:textSize="18sp"
+                    android:visibility="gone" />
+
```
Enkhzul001, line 180: Hmm I think just saying register error would be bit vague for the  person who is registering. Maybe like at each required input there could be error message saying the input is empty or the password and confirm password does not match.

Enkhzul001 (dismissed):
Left few comments. Otherwise, good work!

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
Jacob4649: We should add practice name to the registration UI for providers

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
Jacob4649: It is unnecessary, but we may want to add more cases in the future, in which case we will require either a return or a break

app/src/main/java/com/javahelp/frontend/domain/user/register/IRegisterInput.java
Jacob4649: The backend and frontend should be separate so we cant reuse an interface from the backend in the frontend.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
Lam207: Good idea!

app/src/main/java/com/javahelp/frontend/domain/user/register/RegisterInteractor.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param username     username to register, or null
+     * @param password     {@link byte[]} to register into
+     * @param providerUserInfo user's personal information to register, or null
+     */
+    @Override
+    public void register(String username, String password, ProviderUserInfo providerUserInfo) {
+        if (username == null && providerUserInfo == null) {
+            output.error("Must define one of username, and userinfo");
+        }
+        byte[] salt = SHAPasswordHasher.getInstance().randomSalt();
```
BullDF, line 48: Since you already have an `IPasswordHasher` as an attribute in this class, I think it will be a better idea if we just use `hasher.randomSalt()` instead of getting another instance.

BullDF (approved):
See comments, otherwise, it looks pretty good!

Enkhzul001 (dismissed):
Everything looks good!

Lam207 (commented):
This time only has some not-as-strong suggestions, but let me know if there's something wrong with that.

app/src/main/java/com/javahelp/frontend/activity/ClientRegistrationActivity.java
```diff
@@ -24,16 +47,112 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        clientRegistrationVm.setRegisterError("Registering");
+    }
+
+    /**
+     * Updates this {@link RegisterResult} to reflect the passed register result
+     *
+     * @param registerResult {@link Optional} {@link RegisterResult} to update on
+     */
+    private void updateOnRegisterResult(Optional<RegisterResult> registerResult) {
+        if(registerResult.isPresent()){
```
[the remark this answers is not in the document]
Jacob4649: This suggestion does not compile. Ternary is reserved for expressions that take on different values based on a boolean condition. In this case, no value is being assigned to the result of the ternary expression; and furthermore, both `setVisibility` and `updateOnPresentRegisterResult` are `void` methods and as such, return no values. This is not a situation where ternary expressions can be applied.

app/src/main/java/com/javahelp/frontend/activity/ClientRegistrationVm.java
```diff
@@ -4,10 +4,169 @@
...
+    }
+
+    /**
+     * Tries to Register in
+     */
+    public void attemptRegister(){
+        executor.execute(() ->{
+            try {
+                Thread.sleep(250);
+            }catch (InterruptedException ignored) {
```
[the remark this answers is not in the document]
Jacob4649: The interrupted exception is thrown when something goes wrong with the call to `Thread.sleep(250)` causing it to block for less than 250 milliseconds. In this case, it is a purely visual effect. This exception should never be thrown; however, if it is, it has zero impact on the subsequent code. In this case, it is totally fine to just ignore the exception.

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationActivity.java
Jacob4649: The gateway will send a token back too so there will not be an issue

app/src/main/java/com/javahelp/frontend/activity/ProviderRegistrationViewmodel.java
Jacob4649: I'm not sure I follow how the backend makes this redundant. Remember, we don't want any dependence between our frontend and our backend.

Jacob4649 (approved):

app/src/main/java/com/javahelp/frontend/activity/ClientRegistrationActivity.java
```diff
@@ -24,16 +47,112 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            clientRegistrationVm.setRegisterError("Please enter username and password");
+            return;
+        } else if (!binding.password1.getText().toString().equals(binding.repassword1.getText().toString())) {
+            clientRegistrationVm.setRegisterError("Passwords do not match");
+            return;
+        } else if (binding.email1.getText().toString().isEmpty() || binding.lastname1.getText().toString().isEmpty()
+                || binding.firstname1.getText().toString().isEmpty() || binding.home1.getText().toString().isEmpty()
+                || binding.phonenumber1.getText().toString().isEmpty()) {
+            clientRegistrationVm.setRegisterError("Please enter your personal information");
+            return;
```
Enkhzul001, line 88: great that you implemented warning messages if the form is not filled out

Enkhzul001 (approved):
Everything looks good

Lam207 (approved):
Looks good overall

